PR #11, opened 2017-07-01
New: disallow identical tests

### LINKED ISSUE #5: Rule: disallow identical tests
When a rule has a lot of tests, it's sometimes difficult to tell if any tests are duplicates. This rule would warn if any test cases have the same properties.

Incorrect code for this rule:

```js
/* eslint eslint-plugin/no-identical-tests: error */

ruleTester.run("foo", rule, {
  valid: [
    { code: "foo", options: ["bar"] },
    { code: "somethingElse", options: ["bar"] },
    { code: "foo", options: ["bar"] }
  ],
  invalid: []
});
```

To be honest, I'm not sure how useful this would be in practice -- two test cases might be basically equivalent as far as the rule's logic is concerned, but have slight differences such as whitespace, and `eslint-plugin-eslint-plugin` wouldn't be able to tell that the cases were equivalent.

Aladdin-ADD:
fixes #5

### DIFF
--- a/docs/rules/no-identical-tests.md
+++ b/docs/rules/no-identical-tests.md
@@ -7,7 +7,7 @@ When a rule has a lot of tests, it's sometimes difficult to tell if any tests ar
 Examples of **incorrect** code for this rule:
 
 ```js
-/* eslint eslint-plugin/consistent-output: error */
+/* eslint eslint-plugin/no-identical-tests: error */
 
 new RuleTester().run('foo', bar, {
 valid: [
@@ -22,7 +22,7 @@ invalid: []
 Examples of **correct** code for this rule:
 
 ```js
-/* eslint eslint-plugin/consistent-output: error */
+/* eslint eslint-plugin/no-identical-tests: error */
 
 new RuleTester().run('foo', bar, {
 valid: [
--- a/docs/rules/no-identical-tests.md
+++ b/docs/rules/no-identical-tests.md
@@ -1,0 +1,39 @@
+# Disallow identical tests (no-identical-tests)
+
+When a rule has a lot of tests, it's sometimes difficult to tell if any tests are duplicates. This rule would warn if any test cases have the same properties.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/consistent-output: error */
+
+new RuleTester().run('foo', bar, {
+valid: [
+    { code: 'foo' },
+    { code: 'foo' }
+],
+invalid: []
+});
+
+```
+
+Examples of **correct** code for this rule:
+
+```js
+/* eslint eslint-plugin/consistent-output: error */
+
+new RuleTester().run('foo', bar, {
+valid: [
+    { code: 'foo' },
+    { code: 'bar' }
+],
+invalid: []
+});
+
+```
+
+## When Not To Use It
+
+If you want to allow identical tests, do not enable this rule.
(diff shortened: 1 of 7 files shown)


lib/rules/no-identical-tests.js
```diff
@@ -0,0 +1,53 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'disallow identical tests',
+      category: 'Tests',
+      recommended: true,
```
not-an-aardvark: Can you set this to `false` for now? I agree that it would be good to set it to `true` eventually, but that will be a breaking change, so I think it would be better to set it to `false` at first and then change it a bit afterwards.

lib/rules/no-identical-tests.js
```diff
@@ -0,0 +1,54 @@
...
+    const sourceCode = context.getSourceCode();
+
+    return {
+      Program (ast) {
+        utils.getTestInfo(context, ast).forEach(testRun => {
+          [testRun.valid, testRun.invalid].forEach(tests => {
+            const cache = Object.create(null);
+            // to avoid tests being null
+            (tests || []).forEach(test => {
+              const testCode = sourceCode.getText(test);
```
not-an-aardvark: Eventually, I think it would also be useful to also catch cases like this:

```js
{
  code: 'foo',
  options: ['bar']
},
{
  options: ['bar'],
  code: 'foo'
}
```

But it's probably not necessary to change that as part of this PR.

lib/rules/no-identical-tests.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+    fixable: null,  // or "code" or "whitespace"
+    schema: [],
+  },
+
+  create (context) {
+    // ----------------------------------------------------------------------
+    // Public
+    // ----------------------------------------------------------------------
+    const message = 'This test case should not be identical to someone else.';
```
not-an-aardvark: Nitpick: Can you reword this message to something like this? "This test case is identical to another case."

lib/rules/no-identical-tests.js
Aladdin-ADD: yes, it is a enhancement, I will finish it in another PR.:(

docs/rules/no-identical-tests.md
```diff
@@ -0,0 +1,39 @@
+# Disallow identical tests (no-identical-tests)
+
+When a rule has a lot of tests, it's sometimes difficult to tell if any tests are duplicates. This rule would warn if any test cases have the same properties.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/consistent-output: error */
```
not-an-aardvark: Nitpick: This should say `/* eslint eslint-plugin/no-identical-tests: error */`

docs/rules/no-identical-tests.md
```diff
@@ -0,0 +1,39 @@
...
+],
+invalid: []
+});
+
+```
+
+Examples of **correct** code for this rule:
+
+```js
+/* eslint eslint-plugin/consistent-output: error */
```
not-an-aardvark: Nitpick: This should say `/* eslint eslint-plugin/no-identical-tests: error */`

Aladdin-ADD:
I'm feeling good. @not-an-aardvark

PR #13, opened 2017-07-01
add --fix to no-identical-tests.

Aladdin-ADD:
<del>this is WIP, as the trailing commas have not been handled.</del>

it simply remove the identical tests, so may cause incorrect indent. any suggestions? @not-an-aardvark

### DIFF
--- a/lib/rules/no-identical-tests.js
+++ b/lib/rules/no-identical-tests.js
@@ -42,7 +42,12 @@ module.exports = {
                   node: test,
                   message,
                   fix (fixer) {
-                    return fixer.remove(test);
+                    const start = sourceCode.getTokenBefore(test);
+                    const end = sourceCode.getTokenAfter(test);
+                    return fixer.replaceTextRange(
+                      // should remove test's trailing comma
+                      [start.range[1], end.range[end.value === ',' ? 1 : 0]]
+                      , '');
                   },
                 });
               } else {
--- a/lib/rules/no-identical-tests.js
+++ b/lib/rules/no-identical-tests.js
@@ -46,7 +46,7 @@ module.exports = {
                     const end = sourceCode.getTokenAfter(test);
                     return fixer.replaceTextRange(
                       // should remove test's trailing comma
-                      [start.range[1], end.range[end.value === ',' ? 1 : 0]]
+                      [start.range[1], end.value === ',' ? end.range[1] : test.range[1]]
                       , '');
                   },
                 });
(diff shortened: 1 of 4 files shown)


not-an-aardvark:
Thanks for working on these issues! Sorry about the delay, I'm planning to review your PRs in the next day or two.

> it simply remove the identical tests, so may cause incorrect indent. any suggestions? 

I think that's fine and we don't have to worry about it here -- if the `indent` rule is enabled, the indentation will be automatically fixed.

tests/lib/rules/no-identical-tests.js
```diff
@@ -59,6 +67,16 @@ ruleTester.run('no-identical-tests', rule, {
...
+      output: `
+        new RuleTester().run('foo', bar, {
+          valid: [
+            { code: 'foo' },
+          ],
+          invalid: [
+            { code: 'foo', errors: ['bar'] },
+          ]
+        });
+      `,
```
not-an-aardvark: Can you add a test where there is no trailing comma?

```js
valid: [
  { code: 'foo' }
]
```

lib/rules/no-identical-tests.js
```diff
@@ -41,6 +41,14 @@ module.exports = {
                 context.report({
                   node: test,
                   message,
+                  fix (fixer) {
+                    const start = sourceCode.getTokenBefore(test);
+                    const end = sourceCode.getTokenAfter(test);
+                    return fixer.replaceTextRange(
+                      // should remove test's trailing comma
+                      [start.range[1], end.range[end.value === ',' ? 1 : 0]]
```
not-an-aardvark: I think this will handle whitespace differently depending on whether there is a comma. If there is no comma, then all the whitespace up to the next test will be removed. If there is a comma, the whitespace up to the next test will be preserved.

Maybe it would be better to be consistent about handling whitespace.

Aladdin-ADD:
ping @not-an-aardvark

lib/rules/no-identical-tests.js
```diff
@@ -41,6 +41,14 @@ module.exports = {
                 context.report({
                   node: test,
                   message,
+                  fix (fixer) {
+                    const start = sourceCode.getTokenBefore(test);
+                    const end = sourceCode.getTokenAfter(test);
+                    return fixer.replaceTextRange(
```
not-an-aardvark: Nitpick: This can be `fixer.removeRange` instead of `fixer.replaceTextRange`.

PR #15, opened 2017-07-04
New: fixer-return

### LINKED ISSUE #3: Rule: Always return from a fixer function
```js
/* eslint eslint-plugin/fixer-return: error */

// invalid:

context.report({
  node,
  message,
  fix(fixer) {
    fixer.replaceText(node, "foo");
  }
});

// valid:

context.report({
  node,
  message,
  fix(fixer) {
    return fixer.replaceText(node, "foo");
  }
});

// valid:

context.report({
  node,
  message,
  fix: fixer => fixer.replaceText(node, "foo")
});
```

Aladdin-ADD:
fixes #3

### DIFF
--- a/docs/rules/fixer-return.md
+++ b/docs/rules/fixer-return.md
@@ -1,0 +1,41 @@
+# Enforces always return from a fixer function (fixer-return)
+
+In a fixable rule, missing return from a fixer function will not apply fixes.
+
+## Rule Details
+
+This rule enforces always return from a fixer function.
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/fixer-return: error */
+module.exports = {
+    create: function(context) {
+        context.report( {
+            fix: function(fixer) {
+                fixer.foo();
+            }
+        });
+    }
+};
+```
+
+Examples of **correct** code for this rule:
+
+```js
+/* eslint eslint-plugin/fixer-return: error */
+module.exports = {
+    create: function(context) {
+        context.report( {
+            fix: function(fixer) {
+                return fixer.foo();
+            }
+        });
+    }
+};
+```
+
+## When Not To Use It
+
+If you don't want to enforce always return from a fixer function, do not enable this rule.
--- a/docs/rules/fixer-return.md
+++ b/docs/rules/fixer-return.md
@@ -4,7 +4,7 @@ In a fixable rule, missing return from a fixer function will not apply fixes.
 
 ## Rule Details
 
-This rule enforces always return from a fixer function.
+This rule enforces that fixer functions always return a value.
 
 Examples of **incorrect** code for this rule:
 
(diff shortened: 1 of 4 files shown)


Aladdin-ADD:
@not-an-aardvark sorry about that I'm on a trip, so maybe need a few days to fix this.

not-an-aardvark:
No problem! Enjoy your trip.

lib/rules/fixer-return.js
```diff
@@ -0,0 +1,100 @@
...
+      // Stacks this function's information.
+      onCodePathStart (codePath, node) {
+        const parent = node.parent;
+
+        funcInfo = {
+          upper: funcInfo,
+          codePath,
+          hasReturn: false,
+          shouldCheck:
+            node.type === 'FunctionExpression' &&
```
Aladdin-ADD: we need code path analysis here. so I'm using an inelegant way, rather than using getReportInfo.

am I missing something?

lib/rules/fixer-return.js
```diff
@@ -56,20 +57,27 @@ module.exports = {
...
+      },
 
       // Stacks this function's information.
       onCodePathStart (codePath, node) {
         const parent = node.parent;
+        const shouldCheck = node.type === 'FunctionExpression' &&
+          node.body.type === 'BlockStatement' &&
+          // check if it is context.report({fix})
+          utils.getKeyName(parent) === 'fix' &&
+          parent.parent.type === 'ObjectExpression' &&
```
not-an-aardvark: This will fail for the old-style `context.report` calls:

```js
context.report(node, 'foo', {}, fixer => {});
```

You can use [`utils.getReportInfo`](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/4cec353c272d6dd799ba3c30cd4f06b103b0c7e8/lib/utils.js#L192) to get info about a report.

I think the best way to do it would be to use something like this:

```diff
        const shouldCheck = node.type === 'FunctionExpression' &&
-         node.body.type === 'BlockStatement' &&
-         // check if it is context.report({fix})
-         utils.getKeyName(parent) === 'fix' &&
          parent.parent.type === 'ObjectExpression' &&
          parent.parent.parent.type === 'CallExpression' &&
          contextIdentifiers.has(parent.parent.parent.callee.object) &&
-         parent.parent.parent.callee.property.name === 'report';
+         parent.parent.parent.callee.property.name === 'report' &&
+         utils.getReportInfo(parent.parent.parent.arguments).fix === node;
```

docs/rules/fixer-return.md
```diff
@@ -0,0 +1,41 @@
+# Enforces always return from a fixer function (fixer-return)
+
+In a fixable rule, missing return from a fixer function will not apply fixes.
+
+## Rule Details
+
+This rule enforces always return from a fixer function.
```
not-an-aardvark: Nitpick: I think it would be better to word this as "This rule enforces that fixer functions always return a value."

PR #16, opened 2017-07-04
New: test-case-property-ordering

### LINKED ISSUE #1: Rule: Enforce ordering of keys in test cases
Requires the properties of a test case to be placed in the given order.

```js
/* eslint eslint-plugin/test-case-property-ordering: ["error",
  ["code", "output", "options", "parserOptions", "errors"]
] */

// invalid; wrong order
{
  code: "foo",
  options: ["baz"],
  output: "bar"
}

// valid; correct order
{
  code: "foo",
  output: "bar",
  options: ["baz"]
}

// invalid; extra properties should need to be placed afterwards, I think
{
  code: "foo",
  env: { es6: true }
  output: "bar",
  options: ["baz"],
}

// valid
{
  code: "foo",
  output: "bar",
  options: ["baz"],
  env: { es6: true }
}
```

Aladdin-ADD:
fixes #1

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Name | ✔️ | 🛠 | Description
 [prefer-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-placeholders.md) | | | Disallows template literals as report messages
 [report-message-format](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/report-message-format.md) | | | Enforces a consistent format for report messages
 [require-meta-fixable](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-fixable.md) | ✔️ | | Requires a `meta.fixable` property for fixable rules
-[test-case-property-ordering](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/test-case-property-ordering.md) | | | Requires the properties of a test case to be placed in the given order.
+[test-case-property-ordering](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/test-case-property-ordering.md) | | | Requires the properties of a test case to be placed in a consistent order.
 [test-case-shorthand-strings](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/test-case-shorthand-strings.md) | | 🛠 | Enforces consistent usage of shorthand strings for test cases with no options
 
 ## Supported Presets
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Name | ✔️ | 🛠 | Description
 [prefer-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-placeholders.md) | | | Disallows template literals as report messages
 [report-message-format](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/report-message-format.md) | | | Enforces a consistent format for report messages
 [require-meta-fixable](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-fixable.md) | ✔️ | | Requires a `meta.fixable` property for fixable rules
+[test-case-property-ordering](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/test-case-property-ordering.md) | | | Requires the properties of a test case to be placed in the given order.
 [test-case-shorthand-strings](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/test-case-shorthand-strings.md) | | 🛠 | Enforces consistent usage of shorthand strings for test cases with no options
 
 ## Supported Presets
--- a/docs/rules/test-case-property-ordering.md
+++ b/docs/rules/test-case-property-ordering.md
@@ -2,6 +2,14 @@
 
 ## Rule Details
 
+This rule enforces that the properties of RuleTester test cases are arranged in a consistent order.
+
+### Options
+
+This rule has an array option:
+
+* `["code", "output", "options", "parserOptions", "errors"]` (default): The properties of a test case should be placed in a consistent order.
+
 Examples of **incorrect** code for this rule:
 
 ```js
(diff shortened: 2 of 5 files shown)


docs/rules/test-case-property-ordering.md
```diff
@@ -0,0 +1,55 @@
+# enforce ordering of keys in test cases (test-case-property-ordering)
+
+## Rule Details
```
not-an-aardvark: Can you add a short description of what the rule does? For example:

> This rule enforces that the properties of `RuleTester` test cases are arranged in a consistent order.

docs/rules/test-case-property-ordering.md
```diff
@@ -0,0 +1,55 @@
...
+  output: "bar",
+  options: ["baz"],
+}
+```
+
+### Options
+
+This rule has an array option:
+
+* `["code", "output", "options", "parserOptions", "errors"]` (default): The properties of a test case should be placed in the given order.
```
not-an-aardvark: Can you move this section to above the code samples?

README.md
```diff
@@ -56,6 +56,7 @@ Name | ✔️ | 🛠 | Description
 [prefer-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-placeholders.md) | | | Disallows template literals as report messages
 [report-message-format](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/report-message-format.md) | | | Enforces a consistent format for report messages
 [require-meta-fixable](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-fixable.md) | ✔️ | | Requires a `meta.fixable` property for fixable rules
+[test-case-property-ordering](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/test-case-property-ordering.md) | | | Requires the properties of a test case to be placed in the given order.
```
not-an-aardvark: Nitpick: Can you replace "in the given order" with "in a consistent order"?

lib/rules/test-case-property-ordering.js
```diff
@@ -0,0 +1,62 @@
...
+      type: 'array',
+      elements: { type: 'string' },
+    }],
+  },
+
+  create (context) {
+    // ----------------------------------------------------------------------
+    // Public
+    // ----------------------------------------------------------------------
+    const message = 'The properties of a test case should be placed in the given order: [{{order}}].';
```
not-an-aardvark: Can we change the error message to only contain properties that appear in the test case? For example:

```js
{ output: "baz", code: "foo" }
// -> The properties of a test case should be placed in the given order: [code, output]

{ code: "foo", errors: ["bar"], output: "baz" }
// -> The properties of a test case should be placed in the given order: [code, output, errors]
```

docs/rules/test-case-property-ordering.md
```diff
@@ -0,0 +1,56 @@
...
+```js
+/* eslint eslint-plugin/test-case-property-ordering: ["error",
+  ["code", "output", "options", "parserOptions", "errors"]
+] */
+
+// valid;
+{
+  code: "foo",
+  output: "bar",
+  options: ["baz"],
```
not-an-aardvark: I think these documentation examples are switched around -- the valid samples are in the block that says "Examples of incorrect code", and the invalid samples are in the block that says "Examples of correct code".

lib/rules/test-case-property-ordering.js
```diff
@@ -0,0 +1,62 @@
...
+    // ----------------------------------------------------------------------
+    const message = 'The properties of a test case should be placed in a consistent order: [{{order}}].';
+    const order = context.options[0] || ['code', 'output', 'options', 'parserOptions', 'errors'];
+
+    return {
+      Program (ast) {
+        utils.getTestInfo(context, ast).forEach(testRun => {
+          [testRun.valid, testRun.invalid].forEach(tests => {
+            (tests || []).forEach(test => {
+              const keys = (test.properties || []).map(item => utils.getKeyName(item));
```
not-an-aardvark: Nitpick: This could just be `.map(utils.getKeyName)`, but it doesn't really matter either way.

tests/lib/rules/test-case-property-ordering.js
```diff
@@ -0,0 +1,107 @@
...
+            {
+              code: "foo",
+              env: { es6: true },
+              output: "bar",
+              options: ["baz"],
+            },
+          ]
+        });
+      `,
+      errors: [{ message: 'The properties of a test case should be placed in a consistent order: [code, output, options].' }],
```
not-an-aardvark: I think this error message is a bit misleading because `[code, output, options]` are in the correct order, but the problem is that `env` should be at the end. Maybe it would be better if the error message said `[code, output, options, env]`.

tests/lib/rules/test-case-property-ordering.js
```diff
@@ -0,0 +1,107 @@
...
+              output: "bar",
+              options: ["baz"],
+            },
+          ]
+        });
+      `,
+      options: [['code', 'errors', 'options', 'output', 'parserOptions']],
+      errors: [{ message: 'The properties of a test case should be placed in a consistent order: [code, options, output].' }],
+    },
+  ],
```
not-an-aardvark: Can you add a test where an unknown key is at the beginning?

```js
{ env: {}, code: 'foo' }
```

lib/rules/test-case-property-ordering.js
```diff
@@ -37,18 +37,21 @@ module.exports = {
...
+              const keys = (test.properties || []).map(utils.getKeyName);
 
               for (let i = 0, lastChecked; i < keys.length; i++) {
                 const current = order.indexOf(keys[i]);
 
                 // current < lastChecked to catch unordered;
                 // and lastChecked === -1 to catch extra properties before.
                 if (current > -1 && (current < lastChecked || lastChecked === -1)) {
+                  let orderMsg = order.filter(item => keys.indexOf(item) > -1);
+                  orderMsg = orderMsg.concat(lastChecked === -1 ? [keys[i - 1]] : []);
```
not-an-aardvark: Will this work correctly if `keys` has two unknown keys? It seems like only one of them will be added to `orderMsg`.

Aladdin-ADD: haha..you got it. this is the easiest way and I did want to be a little lazy on this. 😸 😿

Aladdin-ADD: and the extra properties' order is not a must. is it reasonable to give a certain one?

not-an-aardvark: Hmm, that's true. I think it would be fine to say a specific order in the error message, but also allow a different order for the unknown keys.

Aladdin-ADD:
friendly ping. @not-an-aardvark

PR #20, opened 2017-07-16
New: rule prefer-output-null

### LINKED ISSUE #19: Rule: prefer `output: null` to assert that a test case is not autofixed
Disallows invalid `RuleTester` test cases with the output the same as the code.

Invalid:

```js
/* eslint eslint-plugin/prefer-output-null: error */

new RuleTester().run('foo', rule, {
  valid: [],
  invalid: [
    {
      code: 'foo',
      output: 'foo',
      errors: 1
    }
  ]
});
```

Valid:

```js
/* eslint eslint-plugin/prefer-output-null: error */

new RuleTester().run('foo', rule, {
  valid: [],
  invalid: [
    {
      code: 'foo',
      output: null,
      errors: 1
    }
  ]
});
```

```js
/* eslint eslint-plugin/prefer-output-null: error */

new RuleTester().run('foo', rule, {
  valid: [],
  invalid: [
    {
      code: 'foo',
      output: 'bar',
      errors: 1
    }
  ]
});
```

Aladdin-ADD:
fixes #19

### DIFF
--- a/docs/rules/prefer-output-null.md
+++ b/docs/rules/prefer-output-null.md
@@ -1,0 +1,31 @@
+# Disallows invalid RuleTester test cases with the output the same as the code. (prefer-output-null)
+
+## Rule Details
+
+This rule aims to enforce `output: null` for invalid test cases to indicate that no output is produced.
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/prefer-output-null: error */
+
+new RuleTester().run('foo', bar, {
+    valid: [],
+    invalid: [
+      { code: 'foo', errors: [{ message: 'bar' }] },
+    ]
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+/* eslint eslint-plugin/prefer-output-null: error */
+
+new RuleTester().run('foo', bar, {
+    valid: [],
+    invalid: [
+      { code: 'foo', output: null, errors: [{ message: 'bar' }] },
+    ]
+});
+```
--- a/docs/rules/prefer-output-null.md
+++ b/docs/rules/prefer-output-null.md
@@ -2,7 +2,7 @@
 
 ## Rule Details
 
-This rule aims to enforce `output: null` for invalid test cases to indicate that no output is produced.
+The rule reports an error if it encounters a test case where the output is the same as the code.
 
 Examples of **incorrect** code for this rule:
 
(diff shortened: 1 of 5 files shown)


not-an-aardvark:
Oops, I think I might have miscommunicated what the rule should do in https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/19.

I think it's fine for test cases to not have an `output` property at all. For example, some rules aren't fixable, and it would be annoying to specify `output: null` for every test case.

However, the rule should warn when a test case does have an `output` property specified, and the `output` property is the same as the `code` property.

```js
// valid
{
  code: 'foo',
  errors: 1
}

// valid
{
  code: 'foo',
  output: 'bar',
  errors: 1
}

// invalid
{
  code: 'foo',
  output: 'foo', // <-- same as `code`
  errors: 1
}
```

Aladdin-ADD:
Hmm... seem like I missed the tests' meaning. don't worry, easy to modify:(

docs/rules/prefer-output-null.md
```diff
@@ -0,0 +1,31 @@
...
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/prefer-output-null: error */
+
+new RuleTester().run('foo', bar, {
+    valid: [],
+    invalid: [
+      { code: 'foo', errors: [{ message: 'bar' }] },
```
not-an-aardvark: This example should be updated to have `output: 'foo'`.

docs/rules/prefer-output-null.md
```diff
@@ -0,0 +1,31 @@
+# Disallows invalid RuleTester test cases with the output the same as the code. (prefer-output-null)
+
+## Rule Details
+
+This rule aims to enforce `output: null` for invalid test cases to indicate that no output is produced.
```
not-an-aardvark: The behavior of the rule would be clearer if a sentence was added:

> The rule reports an error if it encounters a test case where the output is the same as the code.

lib/rules/prefer-output-null.js
```diff
@@ -0,0 +1,61 @@
...
+    fixable: null,
+    schema: [],
+  },
+
+  create (context) {
+    // ----------------------------------------------------------------------
+    // Public
+    // ----------------------------------------------------------------------
+
+    const message = 'Prefer `output: null` to assert that a test case is not autofixed.';
```
not-an-aardvark: Nitpick: I think it would be better to change the first word to "Use" rather than "Prefer".

PR #21, opened 2017-07-18
Update: no-identical-tests despite of properties order.

### LINKED ISSUE #17: rule changes: no-identical-tests.
check indentical tests despite of property order.

```js
{
  code: 'foo',
  options: ['bar']
},
{
  options: ['bar'],
  code: 'foo'
}
```

expected: no errors.
current: report an error.

Aladdin-ADD:
fixes #17

### DIFF
--- a/lib/rules/no-identical-tests.js
+++ b/lib/rules/no-identical-tests.js
@@ -37,6 +37,14 @@ module.exports = {
      *@returns {boolean} if eq, return true, else return false.
     */
     function eq (ta, tb) {
+      if (ta.type !== tb.type) {
+        return false;
+      }
+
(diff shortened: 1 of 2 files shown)


Aladdin-ADD:
@not-an-aardvark all the tests passed, but linting errors. is this the eslint-plugin-self issue?

not-an-aardvark:
It seems like the issue is that the rule is reporting a lot of false positives on the codebase. I don't think this is related to `eslint-plugin-self`.

Aladdin-ADD:
but when add tests that reporting errors, it works well.

not-an-aardvark:
I think the issue is that it's not accounting for test cases which are just shorthand strings.

PR #25, opened 2017-07-22
Breaking: add some rules to recommended config

Aladdin-ADD:
added these rules:

- no-identical-tests
- <del>prefer-output-null</del>
- fixer-return

these are not added:

- consistent-output
- prefer-placeholders
- report-message-format


not-an-aardvark:
I'm not sure about `prefer-output-null` because some users might have to support an old version of ESLint where `output: null` doesn't work. I like the other suggestions though.

Also, I think these should be `Breaking:` rather than `Update:`.

PR #26, opened 2017-07-22
New: rule no-unused-placeholders.

### LINKED ISSUE #7: Rule: no-unused-placeholders
Reports when a `context.report` call contains a `data` property that does not have a corresponding placeholder in the report message.

Example of invalid code:

```js
context.report({
  node,
  message: "Do not use {{foo}}.",
  data: {
    foo: 1,
    bar: 2
  }
});
```

Aladdin-ADD:
fixes #7

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Name | ✔️ | 🛠 | Description
 [no-deprecated-report-api](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-deprecated-report-api.md) | ✔️ | 🛠 | Prohibits the deprecated `context.report(node, message)` API
 [no-identical-tests](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-identical-tests.md) | ✔️  | 🛠 | Disallows identical tests
 [no-missing-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-missing-placeholders.md) | ✔️ | | Disallows missing placeholders in rule report messages
-[no-unused-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-unused-placeholders.md) | ✔️ | | Disallows unused placeholders in rule report messages
+[no-unused-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-unused-placeholders.md) |  | | Disallows unused placeholders in rule report messages
 [no-useless-token-range](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-useless-token-range.md) | ✔️ | 🛠 | Disallows unnecessary calls to sourceCode.getFirstToken and sourceCode.getLastToken
 [prefer-output-null](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-output-null.md) | | | Disallows invalid RuleTester test cases with the output the same as the code.
 [prefer-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-placeholders.md) | | | Disallows template literals as report messages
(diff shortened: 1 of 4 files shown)


lib/rules/no-unused-placeholders.js
```diff
@@ -0,0 +1,107 @@
...
+// ------------------------------------------------------------------------------
+
+/**
+* Gets report info for a context.report() call
+* TODO: Combine this with logic from no-deprecated-report-api
+* @param {ASTNode[]} reportArgs The arguments to the context.report() call
+* @returns {object}
+*/
+function getReportInfo (reportArgs) {
+  if (reportArgs.length === 1) {
```
not-an-aardvark: Is it possible to use [`utils.getReportInfo`](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/8ac484c68e7314d6977855c1901541f6a3f5dc20/lib/utils.js#L192) for this?

lib/rules/no-unused-placeholders.js
```diff
@@ -0,0 +1,107 @@
...
+            reportInfo &&
+                reportInfo.message &&
+                reportInfo.message.type === 'Literal' &&
+                typeof reportInfo.message.value === 'string' &&
+              reportInfo.data && reportInfo.data.type === 'ObjectExpression'
+          ) {
+            const message = reportInfo.message.value;
+            reportInfo.data.properties.forEach(prop => {
+              const key = utils.getKeyName(prop);
+              const reg = new RegExp(`{{ *${key} *}}`);
```
not-an-aardvark: It's probably not a very common use case, but this will throw an error if `key` contains special characters:

```js
context.report({
  node,
  message: "foo {{ $scope }}"
  data: { $scope: $scope.name }
});
```


Instead of constructing a regular expression for a particular key, I think it would be better to find all used placeholders in the message first (using a regex like [this one](https://github.com/eslint/eslint/blob/2874d75ed8decf363006db25aac2d5f8991bd969/lib/linter.js#L986)), and then check whether each placeholder appears in the object.

tests/lib/rules/no-unused-placeholders.js
```diff
@@ -0,0 +1,128 @@
...
+        module.exports = {
+          create(context) {
+            context.report({
+              node,
+              message: 'foo',
+              data: { bar: '' }
+            });
+          }
+        };
+      `,
```
not-an-aardvark: Can you add a test where a placeholder is used but a different placeholder is specified in `data`?

```js
context.report({
  node,
  message: 'foo {{bar}}',
  data: { baz: '' }
});
```

README.md
```diff
@@ -53,6 +53,7 @@ Name | ✔️ | 🛠 | Description
 [no-deprecated-report-api](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-deprecated-report-api.md) | ✔️ | 🛠 | Prohibits the deprecated `context.report(node, message)` API
 [no-identical-tests](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-identical-tests.md) | ✔️  | 🛠 | Disallows identical tests
 [no-missing-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-missing-placeholders.md) | ✔️ | | Disallows missing placeholders in rule report messages
+[no-unused-placeholders](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-unused-placeholders.md) | ✔️ | | Disallows unused placeholders in rule report messages
```
not-an-aardvark: This line says the rule is recommended, but the rule metadata has `recommended: false`.

(As a separate task, maybe it would be good to autogenerate the table based on rule metadata, so that it doesn't get out of date.)

lib/rules/no-unused-placeholders.js
```diff
@@ -86,12 +49,17 @@ module.exports = {
               reportInfo.data && reportInfo.data.type === 'ObjectExpression'
           ) {
             const message = reportInfo.message.value;
+            // https://github.com/eslint/eslint/blob/2874d75ed8decf363006db25aac2d5f8991bd969/lib/linter.js#L986
+            const PLACEHOLDER_MATCHER = /\{\{\s*([^{}]+?)\s*\}\}/g;
+            const set = new Set();
```
not-an-aardvark: Nitpick: Can you give this variable a more descriptive name? For example: `placeholdersInMessage`

PR #32, opened 2017-07-29
Update: add autofixing to test-case-property-ordering. (fixes #31)

### LINKED ISSUE #31: Add autofixing to test-case-property-ordering
```js
/* eslint eslint-plugin/test-case-property-ordering */

// input:
new RuleTester().run("foo", bar, {
  valid: [
    {
      options: 'bar',
      code: 'foo'
    }
  ]
});

// output:
new RuleTester().run("foo", bar, {
  valid: [
    {
      code: 'foo',
      options: 'bar'
    }
  ]
});
```

Aladdin-ADD:
enable it on eslint codebase: https://github.com/eslint/eslint/pull/9040

my concern is we didn't enable `object-curly-newline`, this may fixing some code like this:
```js
{ code: "var [ x, y ] = z",
options: ["always"],
parserOptions: { ecmaVersion: 6 } },
```

looks so, hmm... strange!

### DIFF
--- a/lib/rules/test-case-property-ordering.js
+++ b/lib/rules/test-case-property-ordering.js
@@ -57,32 +57,10 @@ module.exports = {
                     message,
                     data: { order: orderMsg.join(', ') },
                     fix (fixer) {
-                      const source = sourceCode.getText();
-                      const last = properties[properties.length - 1];
-                      // end loc to replace(including whitespaces).
-                      const finalEnd = sourceCode.getTokenAfter(last, token => token.value === '}' && token.type === 'Punctuator').range[0];
-                      // reorder the properties and put in result array.
-                      const result = [];
-                      for (let j = 0; j < keyNames.length; j++) {
-                        const insertedIndex = orderMsg.indexOf(keyNames[j]);
-                        const propertyCode = sourceCode.getText(properties[j]);
-                        const propertyStart = properties[j].range[1];
-                        const propertyEnd = j < properties.length - 1 ? properties[j + 1].range[0] : finalEnd;
-                        let trailing = source.slice(propertyStart, propertyEnd);
-
-                        // for last property, should check & add trailling commas.
-                        if (j === properties.length - 1 && sourceCode.getTokenAfter(last).value !== ',') {
-                          trailing = ', ' + trailing;
-                        }
-                        result[insertedIndex] = propertyCode + trailing;
-                      }
-
-                      const start = properties[0].range[0];
-
-                      return fixer.replaceTextRange(
-                        [start, finalEnd],
-                        result.join('')
-                      );
+                      return orderMsg.map((key, index) => {
+                        const propertyToInsert = test.properties[keyNames.indexOf(key)];
+                        return fixer.replaceText(test.properties[index], sourceCode.getText(propertyToInsert));
+                      });
                     },
                   });
                 }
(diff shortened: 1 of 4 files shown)


not-an-aardvark:
>my concern is we didn't enable `object-curly-newline`, this may fixing some code like this:
>```js
>{ code: "var [ x, y ] = z",
>options: ["always"],
>parserOptions: { ecmaVersion: 6 } },
>```

Hmm, this doesn't seem ideal. Why not keep the old spacing, and just swap the properties around? That way, comments will also be preserved.

Aladdin-ADD:
ping @not-an-aardvark

lib/rules/test-case-property-ordering.js
```diff
@@ -31,29 +31,59 @@ module.exports = {
...
+                      const result = [];
+                      for (let j = 0; j < keyNames.length; j++) {
+                        const insertedIndex = orderMsg.indexOf(keyNames[j]);
+                        const propertyCode = sourceCode.getText(properties[j]);
+                        const propertyStart = properties[j].range[1];
+                        const propertyEnd = j < properties.length - 1 ? properties[j + 1].range[0] : finalEnd;
+                        let trailing = source.slice(propertyStart, propertyEnd);
+
+                        // for last property, should check & add trailling commas.
+                        if (j === properties.length - 1 && sourceCode.getTokenAfter(last).value !== ',') {
```
not-an-aardvark: Instead of having a special case for trailing commas, would it work to just set `finalEnd` to the trailing comma if it exists?

```js
const finalEnd = sourceCode.getTokenAfter(last);
```

lib/rules/test-case-property-ordering.js
```diff
@@ -31,29 +31,59 @@ module.exports = {
...
+                      const last = properties[properties.length - 1];
+                      // end loc to replace(including whitespaces).
+                      const finalEnd = sourceCode.getTokenAfter(last, token => token.value === '}' && token.type === 'Punctuator').range[0];
+                      // reorder the properties and put in result array.
+                      const result = [];
+                      for (let j = 0; j < keyNames.length; j++) {
+                        const insertedIndex = orderMsg.indexOf(keyNames[j]);
+                        const propertyCode = sourceCode.getText(properties[j]);
+                        const propertyStart = properties[j].range[1];
+                        const propertyEnd = j < properties.length - 1 ? properties[j + 1].range[0] : finalEnd;
```
not-an-aardvark: I think these variable names are confusing because `propertyStart` and `propertyEnd` describe the boundaries of the whitespace after the property, not the property itself. Maybe `whitespaceStart` and `whitespaceEnd` would be better.

lib/rules/test-case-property-ordering.js
```diff
@@ -31,29 +31,59 @@ module.exports = {
...
+                        result[insertedIndex] = propertyCode + trailing;
+                      }
+
+                      const start = properties[0].range[0];
+
+                      return fixer.replaceTextRange(
+                        [start, finalEnd],
+                        result.join('')
+                      );
+                    },
```
not-an-aardvark: Would it be possible to use the [new feature in ESLint 4.1.0](http://eslint.org/blog/2017/06/eslint-v4.1.0-released#applying-multiple-autofixes-simultaneously) where you can return an array of autofixes?

That way, this autofix would be very simple, because it wouldn't be necessary to deal with whitespace or trailing commas:

```js
fix(fixer) {
  return orderMsg.map((key, index) => {
    const propertyToInsert = test.properties[keyNames.indexOf(key)];
    return fixer.replaceText(test.properties[index], sourceCode.getText(propertyToInsert));
  });
}
```

Aladdin-ADD: awesome!!!

PR #41, opened 2017-09-09
New: no-deprecated-context-methods rule (fixes #40)

### LINKED ISSUE #40: Rule: avoid deprecated `context` passthroughs
Disallows rules from using the following deprecated methods on `context`:

* `getSource`
* `getSourceLines`
* `getAllComments`
* `getNodeByRangeIndex`
* `getComments`
* `getCommentsBefore`
* `getCommentsAfter`
* `getCommentsInside`
* `getJSDocComment`
* `getFirstToken`
* `getFirstTokens`
* `getLastToken`
* `getLastTokens`
* `getTokenAfter`
* `getTokenBefore`
* `getTokenByRangeStart`
* `getTokens`
* `getTokensAfter`
* `getTokensBefore`
* `getTokensBetween`

not-an-aardvark:


Aladdin-ADD:
I think this can be added to reccommended, since we have a peerdependence eslint > 4.1

not-an-aardvark:
I agree. Although now that we've released v1.0.0, we would need to release v2.0.0 to add it to `recommended`. It would be nice to not have too many breaking changes.

Maybe we should leave it for now, and then plan to add it to `recommended` next time we do a major release.

Aladdin-ADD:
agree! and prefer-output-null for the same reason.

PR #50, opened 2017-10-22
New: rule prefer-replace-text

Aladdin-ADD:


docs/rules/prefer-replace-range.md
```diff
@@ -0,0 +1,52 @@
...
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/prefer-text-range: error */
+module.exports = {
+  create(context) {
+    context.report({
+      fix(fixer) {
+        // error, can be writen: return fixer.replaceText([node, '');
```
not-an-aardvark: Nitpick: the word "written" is spelled incorrectly here. Also, the comment here is missing a `]`.

lib/rules/prefer-replace-text.js
```diff
@@ -0,0 +1,83 @@
...
+      category: 'Rules',
+      recommended: false,
+    },
+    fixable: null,
+    schema: [],
+  },
+
+  create (context) {
+    const sourceCode = context.getSourceCode();
+    const message = 'prefer using replaceText instead of replaceTextRange.';
```
not-an-aardvark: Nitpick: I think it would be better to make the message:

```js
const message = 'Use replaceText instead of replaceTextRange.';
```

PR #54, opened 2018-01-08
Add URL to rule documentation to the metadata

Arcanemagus:
ESLint v4.15.0 added an official location for rules to store a URL to their documentation in the rule metadata in eslint/eslint#9788. This adds the URL to all the existing rules so anything consuming them can
know where their documentation is without having to resort to external packages to guess.

### DIFF
--- a/lib/rules/consistent-output.js
+++ b/lib/rules/consistent-output.js
@@ -17,6 +17,7 @@ module.exports = {
       description: 'Enforce consistent use of output assertions in rule tests',
       category: 'Tests',
       recommended: false,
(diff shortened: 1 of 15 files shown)


not-an-aardvark:
Thanks for the PR! I decided to do this a slightly different way in https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/commit/eb207d0ef650e9f91f7f7ca3be2ab8cbe762f0d3 so that the URLs are versioned/dynamically generated.

Arcanemagus:
Great! I was just going through [the list](https://github.com/jfmengels/eslint-rule-documentation/blob/master/plugins.json) that `eslint-rule-documentation` supported 😛.

Still need to get support in the `linter-eslint` plugin for Atom lol (https://github.com/AtomLinter/linter-eslint/pull/1067).

Issue #60, opened 2018-07-24
Require valid rule categories


not-an-aardvark:
This rule would require that all lint rules have a `meta.docs.category` property which is one of several user-specified strings. The goal would be to prevent mistakes like the one pointed out in https://github.com/eslint/eslint/pull/10661#discussion_r204618870.

Invalid example:

```js
/* eslint eslint-plugin/require-valid-categories: [error, ["Possible Errors", "Stylistic Issues"]] */

module.exports = {
    meta: {
        docs: {
            category: "Not a valid category"
        }
    },
    create(context) {
        // ...
    }
};
```

Valid example:

```js
/* eslint eslint-plugin/require-valid-categories: [error, ["Possible Errors", "Stylistic Issues"]] */

module.exports = {
    meta: {
        docs: {
            category: "Possible Errors"
        }
    },
    create(context) {
        // ...
    }
};
```

PR #63, opened 2018-09-24
WIP: Add require-valid-categories rule

macklinu:
See #60 

This is a start at implementing the rule in #60. Still a WIP, but wanted to open the PR to start a discussion and also let you know that someone is working on it. 🎉 

TODO: 

- [ ] add rule category
- [ ] add proper JSDoc
- [ ] add more test cases (old rule format, new rule format, empty category option, no category option, invalid category option, like numbers)
- [ ] add `docs/rules/require-valid-categories.md`

### DIFF
--- a/lib/rules/require-valid-categories.js
+++ b/lib/rules/require-valid-categories.js
@@ -1,0 +1,57 @@
+'use strict';
+
+const utils = require('../utils');
+
+module.exports = {
+  meta: {
+    docs: {},
+    category: '', // fixme: add valid category
+    schema: {
+      type: 'array',
+      items: { type: 'string' },
+      minItems: 1,
+      uniqueItems: true,
+    },
+  },
+  create (context) {
+    const options = new Set(context.options || []);
+
+    // fixme: add JSDoc
+    function isValidCategory (node) {
+      return Boolean(
+        node &&
+        node.type === 'Literal' &&
+        typeof node.value === 'string' &&
+        options.has(node.value)
+      );
+    }
+
+    return {
+      Program (node) {
+        const info = utils.getRuleInfo(node);
+        if (!info) {
+          return;
+        }
+
+        const metaNode = info.meta;
+        const docsPropNode =
+          metaNode &&
+          metaNode.properties &&
+          metaNode.properties.find(p => p.type === 'Property' && utils.getKeyName(p) === 'docs');
+        const categoryPropNode =
+          docsPropNode &&
+          docsPropNode.value.properties &&
+          docsPropNode.value.properties.find(p => p.type === 'Property' && utils.getKeyName(p) === 'category');
+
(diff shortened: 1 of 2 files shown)


lib/rules/require-valid-categories.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+const utils = require('../utils');
+
+module.exports = {
+  meta: {
+    docs: {},
+    category: '', // fixme: add valid category
+    schema: {
```
macklinu: My thought is that this rule should be opt-in only and requires options (otherwise what categories would you validate against)? As such, the only option this rule takes is an array of strings containing at least one value. Otherwise, ESLint should report that the rule was configured improperly. Thoughts on this?

lib/rules/require-valid-categories.js
```diff
@@ -0,0 +1,57 @@
...
+          docsPropNode.value.properties &&
+          docsPropNode.value.properties.find(p => p.type === 'Property' && utils.getKeyName(p) === 'category');
+
+        if (isValidCategory(categoryPropNode && categoryPropNode.value)) {
+          return;
+        }
+
+        context.report({
+          node: categoryPropNode.value,
+          message: 'Not a valid category.',
```
macklinu: Probably could use a better message too. 🤔

lib/rules/require-valid-categories.js
not-an-aardvark: My only concern here is that this plugin has the `plugin:eslint-plugin/all` config, which currently claims to enable all rules from the plugin. This isn't possible unless all rules can be run without providing an explicit configuration. So my slight preference would be to have some default set of categories (e.g. `Possible Errors`, `Best Practices`, `Stylistic Issues`) if the user doesn't provide any categories on their own.

lib/rules/require-valid-categories.js
not-an-aardvark: Agreed -- maybe add the names of the valid categories? For example: `Category should be one of 'Foo', 'Bar', or 'Baz'.`

aladdin-add:
@armano2  the rule was configurable,  as described in the issue #60 😄 
> This rule would require that all lint rules have a meta.docs.category property which is one of several user-specified strings. 

friendly ping @macklinu are you still working on this?

macklinu:
@aladdin-add sorry for dropping the ball on this (missed the message in my email inbox). Will pick back up on this rule this week. 😄

lib/rules/require-valid-categories.js
j-f1: Perhaps all the categories ESLint core uses? It would be nice to be able to use those and combine them with plugin-specific options.

aladdin-add:
friendly ping @macklinu

aladdin-add:
closing, as the `category` will be removed in eslint v8. :)

PR #68, opened 2018-12-04
New: require-meta-type (fixes #67)

### LINKED ISSUE #67: new rule to enforce meta.type
https://eslint.org/blog/2018/11/eslint-v5.9.0-released#the-fix-type-option

Aladdin-ADD:


lib/rules/require-meta-type.js
```diff
@@ -0,0 +1,66 @@
...
+    docs: {
+      description: 'require rules to implement a meta.type property',
+      category: 'Rules',
+      recommended: false, // TODO: enable it in a major release.
+    },
+    type: 'problem',
+    fixable: null,
+    schema: [],
+    messages: {
+      missing: '`meta.type` is required.',
```
not-an-aardvark: If someone is unfamiliar with `meta.type`, it might be useful to list the possible values in this message too. For example:

> \`meta.type\` is required (must be either \`problem\`, \`suggestion\`, or \`layout\`).

lib/rules/require-meta-type.js
```diff
@@ -0,0 +1,66 @@
...
+        const metaNode = info.meta;
+        const typeNode =
+          metaNode &&
+          metaNode.properties &&
+          metaNode.properties.find(p => p.type === 'Property' && utils.getKeyName(p) === 'type');
+
+        // if has no typeNode => report "missing"
+        // if has invalid value => report "unexpected"
+        const messageId = typeNode
+          ? validTypes.indexOf(typeNode.value.value) < 0 ? 'unexpected' : ''
```
not-an-aardvark: This ternary expression seems a bit confusing. Could you separate this onto multiple lines?

```js
if (typeNode && typeNode.value.type === 'Literal' && validTypes.includes(typeNode.value.value)) {
    context.report({ node: metaNode, messageId: 'unexpected' });
} else if (!typeNode) {
    context.report({ node: metaNode, messageId: 'missing' });
}
```

lib/rules/require-meta-type.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * @fileoverview require rules to implement a meta.type property
+ * @author 薛定谔的猫<weiran.zsd@outlook.com>
+ */
+
+'use strict';
+
+const utils = require('../utils');
+const validTypes = ['problem', 'suggestion', 'layout'];
```
not-an-aardvark: Nitpick: Can you capitalize this name and make it a `Set`?

```js
const VALID_TYPES = new Set(['problem', 'suggestion', 'layout']);
```

PR #75, opened 2018-12-16
Fix: allow to use generator function as fix in fixer-return

armano2:


not-an-aardvark:
Released as v2.0.1, thanks!

PR #77, opened 2019-02-08
Update: add plugins in configs

aladdin-add:
it aims to simplify usage of the plugin (so users no need to specify it).
but the tests failed, as eslint cannot find `eslint-plugin-eslint-plugin` when linting the repo. any ideas?


not-an-aardvark:
Sorry, I lost track of this.

Personally, I don't think it's necessary to do this, it's not very difficult for users to add `plugins` themselves. But I'm not opposed if we can figure out a good way to get the tests passing.

PR #80, opened 2019-04-12
New: meta-property-ordering (fixes #62)

### LINKED ISSUE #62: New: meta-property-ordering
similar to rule test-case-property-ordering, but it aims to check the ordering rule meta.

not sure if it can enforce the exsistence of some properties (e.g. fixalbe, docs, messages), or another rule?

golopot:
fixes #62

### DIFF
--- a/docs/rules/meta-property-ordering.md
+++ b/docs/rules/meta-property-ordering.md
@@ -2,7 +2,7 @@
 
 (fixable) The `--fix` option on the [command line](../user-guide/command-line-interface#fix) automatically fixes problems reported by this rule.
 
-This rule enforces that the properties of rule meta are arranged in a consistent order.
+This rule enforces that meta properties of a rule are placed in a consistent order.
 
 ## Rule Details
 
@@ -10,7 +10,7 @@ This rule enforces that the properties of rule meta are arranged in a consistent
 
 This rule has an array option:
 
-* `['type', 'docs', 'fixable', 'schema', 'messages', 'deprecated', 'replacedBy']` (default): The properties of meta should be placed in a consistent order.
+* `['type', 'docs', 'fixable', 'schema', 'messages', 'deprecated', 'replacedBy']` (default): The order that the properties of `meta` should be placed in.
 
 Examples of **incorrect** code for this rule:
 
@@ -20,13 +20,26 @@ Examples of **incorrect** code for this rule:
   ["type", "docs", "fixable", "schema", "messages"]
 ] */
 
-// invalid; wrong order
-{
-  fixable: false,
-  type: "problem",
-  docs: "",
+// invalid; wrong order.
+module.exports = {
+  meta: {
+    docs: "",
+    type: "problem",
+    fixable: "code",
+  },
+  create() {},
 }
 
+// invalid; extra properties must be placed afterwards.
+module.exports = {
+  meta: {
+    type: "problem",
+    fooooooooo: "foo",
+    docs: "",
+    fixable: "code",
+  },
+  create() {},
+}
 ```
 
 Examples of **correct** code for this rule:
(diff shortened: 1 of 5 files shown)


golopot:
Needs to decide the default ordering. After running the rule against the eslint code base. I find that the order `[type, docs, fixable, schema, messages, deprecated, replacedBy]` is the best match.

Currently this implementation does not care about the ordering of unknown meta properties.

docs/rules/meta-property-ordering.md
```diff
@@ -0,0 +1,51 @@
...
+```js
+/* eslint eslint-plugin/test-case-property-ordering: ["error",
+  ["type", "docs", "fixable", "schema", "messages"]
+] */
+
+// valid;
+{
+  type: "bar",
+  docs: "foo",
+  fooooooooo: "foo",
```
aladdin-add: To be consistent with the similar rule [test-case-property-ordering](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/test-case-property-ordering.md), extra properties should need to be placed afterwards. What do you think?

lib/rules/meta-property-ordering.js
```diff
@@ -0,0 +1,78 @@
...
+          !info.meta ||
+          info.meta.properties.length < 2
+        ) {
+          return;
+        }
+
+        const propsActual = info.meta.properties
+          .filter(prop => orderMap.has(getKeyName(prop)));
+
+        for (let i = 1, j = propsActual.length; i < j; i += 1) {
```
aladdin-add: nits: can be `i < propsActual.length`?
 is `j` used somewhere?

golopot: I saw this pattern multiple times in the codebase of `eslint-plugin-react`. I think people write this for a potential performance gain. Since I don't know the truth about this, I will change it as requested.

docs/rules/meta-property-ordering.md
golopot: Yes.

docs/rules/meta-property-ordering.md
```diff
@@ -0,0 +1,51 @@
...
+
+This rule enforces that the properties of rule meta are arranged in a consistent order.
+
+## Rule Details
+
+### Options
+
+This rule has an array option:
+
+* `['type', 'docs', 'fixable', 'schema', 'messages', 'deprecated', 'replacedBy']` (default): The properties of meta should be placed in a consistent order.
```
not-an-aardvark:
```suggestion
* `['type', 'docs', 'fixable', 'schema', 'messages', 'deprecated', 'replacedBy']` (default): The order that the properties of `meta` should be placed in.
```

docs/rules/meta-property-ordering.md
```diff
@@ -0,0 +1,51 @@
...
+Examples of **incorrect** code for this rule:
+
+```js
+
+/* eslint eslint-plugin/meta-property-ordering: ["error",
+  ["type", "docs", "fixable", "schema", "messages"]
+] */
+
+// invalid; wrong order
+{
```
not-an-aardvark: Can you update these examples to include a full rule with the `meta` property, rather than just being an object?

docs/rules/meta-property-ordering.md
golopot: Updated as requested.

lib/rules/meta-property-ordering.js
golopot: This is outdated after last rewrite.

docs/rules/meta-property-ordering.md
golopot: updated as suggested.

docs/rules/meta-property-ordering.md
golopot: updated as suggested.

tests/lib/rules/meta-property-ordering.js
```diff
@@ -0,0 +1,133 @@
...
+        };`,
+
+      output: `
+        module.exports = {
+          meta: {type, fooooooooo, doc, fixable},
+          create() {},
+        };`,
+      options: [['type', 'doc', 'fixable']],
+      errors: [
+        { message: getMessage(['type', 'doc', 'fixable']) },
```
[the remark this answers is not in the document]
golopot: Since the updated implementation places unknown properties at the end, and that the message behaves just like `test-case-property-ordering`, are you still interesting in this change?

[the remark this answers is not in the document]
aladdin-add: I'd slightly like it, but it can be in another PR(as an enhancement ). 😄

not-an-aardvark:
Published in v2.1.0.

PR #84, opened 2019-07-28
Fix: Check for meta type even when using a function reference (fixes #83)

### LINKED ISSUE #83: require-meta-type does not work if `create` is not a function
The following does not fail linting:

```
/* eslint eslint-plugin/require-meta-type: error */

function create(context) {
	// ...
}

module.exports = {
	meta: {}, // should fail
	create
};
```

----

```
/* eslint eslint-plugin/require-meta-type: error */

const create = (context) => {
	// ...
}

module.exports = {
	meta: {}, // should fail
	create
};
```

----

It looks like the utility for checking if a file contains a rule has a [hard check](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/lib/utils.js#L93) for `create` being a function. Is this check necessary? Can I submit a PR to either:

* Allow identifiers on the right side of `create` even if they are not functions?
* Or follow identifiers on the right side of `create` if they reference a function in the same file?

MrHen:
Fixes #83.

### DIFF
--- a/lib/rules/require-meta-type.js
+++ b/lib/rules/require-meta-type.js
@@ -30,7 +30,7 @@ module.exports = {
 
   create (context) {
     const sourceCode = context.getSourceCode();
-    const info = utils.getRuleInfo(sourceCode.ast);
+    const info = utils.getRuleInfo(sourceCode.ast, sourceCode.scopeManager);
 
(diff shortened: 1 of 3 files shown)


aladdin-add:
friendly ping @not-an-aardvark

PR #89, opened 2019-12-05
Add new rule `require-meta-docs-description`

bmish:
The eslint core plugin has an internal lint rule to enforce this same thing: https://github.com/eslint/eslint/blob/master/tools/internal-rules/consistent-docs-description.js


bmish:
Hey @not-an-aardvark, I've got several PRs open in this repository, for improvements I'd like to make use of in eslint-plugin-ember which I maintain, and which I'd also like to use in the core eslint plugin to replace some of the custom linting they have.

docs/rules/require-meta-docs-description.md
```diff
@@ -0,0 +1,41 @@
+# require rules to implement a meta.docs.description property (require-meta-docs-description)
+
+Defining a clear and consistent description for each rule helps developers understand what they're used for.
+
+In particular, each rule description should begin with an allowed prefix:
+* `enforce`
+* `require`
+* `disallow`
```
not-an-aardvark: I think this one should probably be configurable with something like a regex (maybe with `enforce`/`require`/`disallow` as the default). As far as I know, `enforce`/`require`/`disallow` is a convention used by the ESLint team but it's not universally adopted.

bmish: Good point, I'll update shortly.

bmish: I added an `allowedPrefixes` array option to the rule.

bmish:
@not-an-aardvark it seems like an unrelated commit beafe713387f7ef2a6ff844400e231bbbe73c5fb was accidentally added to this PR? That commit was already merged in a separate PR.

not-an-aardvark:
Sorry, that's probably my fault (I force-pushed to master to update a commit message.)

bmish:
Okay I just rebased to fix it. 

Let me know what you think of the PR and then I'd be interested to get a new release once this is merged.

bmish:
@not-an-aardvark ping?

docs/rules/require-meta-docs-description.md
not-an-aardvark: I'd really rather this be a regex option -- choosing a set of specific prefixes feels like an oddly specific restriction to impose on users of the rule.

bmish: @not-an-aardvark sounds good, I switched to a `pattern` option.

bmish:
Thanks for the review @not-an-aardvark! Can we get a release?

not-an-aardvark:
Published as v2.2.0, thanks for your patience

PR #90, opened 2020-01-16
Update `require-meta-schema` rule to allow object schemas (in addition to array schemas)

bmish:
Turns out that eslint supports both array and object schemas. eslint itself has a handful of rules that use objects schemas, although array schemas are much more common.

https://eslint.org/docs/developer-guide/working-with-rules#options-schemas

Example rule with object schema: https://github.com/eslint/eslint/blob/master/lib/rules/eqeqeq.js#L29
Example rule with array schema: https://github.com/eslint/eslint/blob/master/lib/rules/func-style.js#L22


not-an-aardvark:
Thanks, published v2.2.1

PR #91, opened 2020-04-05
add the plugin name to plugins prop of presets

Gerhut:
This will make no needs to add `eslint-plugin` to `plugins` manually if a preset is enabled.


aladdin-add:
not sure why the build was not triggered. / cc @not-an-aardvark

bmish:
Can we merge this? This is actually a good/important improvement.

Gerhut:
rebased

PR #98, opened 2020-06-06
Fix: ensure that node has `key` property

G-Rath:
Found while writing tests for [eslint-plugin-eslint-config](https://github.com/G-Rath/eslint-plugin-eslint-config)


not-an-aardvark:
Thanks for the PR. I ended up merging https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/100 which also fixes this issue.

PR #100, opened 2020-06-09
Fix: handle spreads in rule meta objects

### LINKED ISSUE #99: Cannot read property 'type' of undefined
I have an "extension rule" which spreads the base rule to modify the error message.

## Rule source

```js
'use strict';

const baseRule = require('eslint/lib/rules/no-console');

module.exports = {
  ...baseRule,
  meta: {
    ...baseRule.meta,
    messages: {
      unexpected: `My extended error message`,
    },
  },
};
```

## Expected

Linter runs fine

## Actual

```
TypeError: Error while loading rule 'eslint-plugin/require-meta-fixable': Cannot read property 'type' of undefined
      Occurred while linting <my rule>.js
          at Object.getKeyName (<snip>/node_modules/eslint-plugin-eslint-plugin/lib/utils.js:187:44)
          at node.right.properties.reduce (<snip>/node_modules/eslint-plugin-eslint-plugin/lib/utils.js:114:47)
          at Array.reduce (<anonymous>)
          at ast.body.filter.map.filter.filter.reduce (<snip>/node_modules/eslint-plugin-eslint-plugin/lib/utils.js:113:42)
          at Array.reduce (<anonymous>)
          at Object.getRuleInfo (<snip>/eslint-plugin-eslint-plugin/lib/utils.js:97:8)
```

The problem is here; the function assumes the property is a `Property`, but in this case it is a `SpreadElement`.

https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/68059b1ed2cb2710a4d6c4aa1d05afad8de565f0/lib/utils.js#L187

bradzacher:
Fixes #99

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -184,6 +184,10 @@ module.exports = {
   * @returns {string|null} The key name, or `null` if the name cannot be determined statically.
   */
   getKeyName (property) {
+    if (!property.key) {
+      // likely a SpreadElement or another non-standard node
+      return null;
+    }
     if (!property.computed && property.key.type === 'Identifier') {
       return property.key.name;
     }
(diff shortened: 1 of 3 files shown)


not-an-aardvark:
Thanks, released in v2.2.2.

PR #105, opened 2021-04-07
New: Add new rule `require-meta-has-suggestions`

bmish:
Suggestable ESLint rules should have a `meta.hasSuggestions` property to indicate that they provide [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions). This new rule enforces that the `meta.hasSuggestions` property is correctly enabled when a rule provides suggestions, and not enabled when a rule does not provide suggestions.
    
The change to require suggestable rules to have `meta.hasSuggestions` has been accepted and mentioned in the blog post for the upcoming ESLint 8 breaking changes. So we should adopt this change now to help plugin authors ensure they are compatible with ESLint 8 as soon as possible. The old property `meta.docs.suggestion` was unused anyway.
    
https://eslint.org/blog/2021/06/whats-coming-in-eslint-8.0.0#rules-with-suggestions-now-require-the-metahassuggestions-property

We should make this a `recommended` rule in the next major release.
    
This is very similar to the existing [eslint-plugin/require-meta-fixable](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-fixable.md) rule enforcing the correct presence of the `meta.fixable` property.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Name | ✔️ | 🛠 | Description
 [require-meta-docs-description](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-docs-description.md) |  |  | require rules to implement a meta.docs.description property with the correct format
 [require-meta-docs-url](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-docs-url.md) |  | 🛠 | require rules to implement a meta.docs.url property
 [require-meta-fixable](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-fixable.md) | ✔️ |  | require rules to implement a meta.fixable property
+[require-meta-has-suggestions](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-has-suggestions.md) |  |  | require suggestable rules to implement a `meta.hasSuggestions` property
 [require-meta-schema](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-schema.md) |  | 🛠 | require rules to implement a meta.schema property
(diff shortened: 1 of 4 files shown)


lib/rules/require-meta-docs-suggestion.js
```diff
@@ -0,0 +1,73 @@
...
+    const ruleInfo = utils.getRuleInfo(sourceCode.ast);
+    let contextIdentifiers;
+    let ruleReportsSuggestions;
+
+    return {
+      Program (node) {
+        contextIdentifiers = utils.getContextIdentifiers(context, node);
+      },
+      CallExpression (node) {
+        if (
```
aladdin-add, line 35: it could be simple to use the util: https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/a3017e253db08630da3f92428df6f4f2680ecf12/lib/utils.js#L209 and [selectors](https://eslint.org/docs/developer-guide/selectors#selectors). :)

aladdin-add (changes requested):
just a suggestion, but otherwise lgtm, thanks!

bmish:
I have moved this PR to draft status pending the outcome of this issue https://github.com/eslint/eslint/issues/14312 to discuss the future of the `meta.docs.suggestion` property.

lib/rules/require-meta-docs-suggestion.js
bmish: Will look into this.

bmish: I'm not sure if it's convenient to use selectors here since we have a variable `contextIdentifiers` in the condition.

bmish:
Now that the `meta.hasSuggestions` property has been [accepted](https://eslint.org/blog/2021/06/whats-coming-in-eslint-8.0.0#rules-with-suggestions-now-require-the-metahassuggestions-property) as part of ESLint 8, I have updated the rule to target that property, and we can merge/release this anytime.

aladdin-add (approved):
LGTM, thanks! 💯

not-an-aardvark (approved):

Issue #104, opened 2021-04-07
switch to github actions


aladdin-add:
travis was not triggered in recent changes. maybe it's time to drop it, and switch to github actions. :)

bradzacher:
Would recommend.
When we switched Typescript-ESLint to actions we had a lot less issues.

fisker:
I'll send PR.

PR #107, opened 2021-04-09
Fix: Rule can't find reference of `create` function

### LINKED ISSUE #106: `getRuleInfo` not working
[`getRuleInfo`](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/02c1f347f2f0881e364724a4fce69e9da2ce9313/lib/utils.js#L87) seems not working if rule write in this way

```js
'use strict';

const create = () => {};

module.exports = {create};
```

See https://github.com/sindresorhus/eslint-plugin-unicorn/blob/e7e25b82526e42970043c39a512cbdfb889b7cb7/rules/better-regex.js#L107

I dig a little bit, [this line](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/02c1f347f2f0881e364724a4fce69e9da2ce9313/lib/utils.js#L32) , `scopeManager.acquire(node)` is null, so it will use `globalScope`. If I change to `scopeManager.globalScope.childScopes[0]` it will work.

fisker:
Turns out if `create` is a reference, the global scope don't have the reference not only in `nodejsScope`, also broken if `sourceType` is set to `'module'`.

To find the correct reference, I'm searching all `childScopes`, I hope it's fine to do this.

Fixes #106

### DIFF
--- a/tests/lib/utils.js
+++ b/tests/lib/utils.js
@@ -122,6 +122,31 @@ describe('utils', () => {
           );
         });
       });
+
+      for (const scopeOptions of [
+        { ignoreEval: true, ecmaVersion: 6, sourceType: 'script', nodejsScope: true },
+        { ignoreEval: true, ecmaVersion: 6, sourceType: 'script' },
+        { ignoreEval: true, ecmaVersion: 6, sourceType: 'module' },
+      ]) {
+        const ast = espree.parse(`
+          const create = () => {};
+          const meta = {};
+          module.exports = { create, meta };
+        `, { ecmaVersion: 6 });
+        const expected = {
+          create: { type: 'Identifier' },
+          meta: { type: 'Identifier' },
+          isNewStyle: true,
+        };
+        it(`ScopeOptions: ${JSON.stringify(scopeOptions)}`, () => {
+          const scope = escope.analyze(ast, scopeOptions);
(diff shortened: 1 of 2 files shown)


fisker:
@aladdin-add  No CI running?

tests/lib/utils.js
```diff
@@ -122,6 +122,31 @@ describe('utils', () => {
...
+        { ignoreEval: true, ecmaVersion: 6, sourceType: 'module' },
+      ]) {
+        const ast = espree.parse(`
+          const create = () => {};
+          const meta = {};
+          module.exports = { create, meta };
+        `, { ecmaVersion: 6 });
+        const expected = {
+          create: { type: 'Identifier' },
+          meta: { type: 'Identifier' },
```
fisker, line 138: This should still break rules like [this](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/02c1f347f2f0881e364724a4fce69e9da2ce9313/lib/rules/meta-property-ordering.js#L42), It should return the reference. But I'm not going to work on it in this PR.

aladdin-add:
ci was not working somehow ╮(￣▽￣)╭, we are considering switching to gh actions. refs #104

aladdin-add (approved):
LGTM, thanks!

not-an-aardvark (approved):

PR #108, opened 2021-04-09
Switch to github actions

fisker:
Fixes #104

PR #109, opened 2021-04-09
Chore: Switch `escope` to `eslint-scope`

fisker:
`escope` is old, and `eslint-scope` is eslint using.


package.json
```diff
@@ -7,7 +7,6 @@
   "license": "MIT",
   "scripts": {
     "lint": "eslint . --ignore-pattern \"!.*\"",
-    "pretest": "npm run lint",
```
aladdin-add, line 10: why removing this?

fisker: When I send #108, I didn't notice, run `test` also run `lint`, so I setup a separate step for `lint` https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/31951d4127f2b49cda7400a41629c05091a1ea9d/.github/workflows/main.yml#L30

So `lint` will run twice, you prefer the old way or remove the `lint` step on Github Actions?

aladdin-add: I see. no need to lint twice. 👍

aladdin-add (approved):

PR #111, opened 2021-04-15
`require-meta-schema`: Fix false positive

### LINKED ISSUE #110: `require-meta-schema`: False positive
When schema is a reference, `require-meta-schema` reported on following case.

```js
const schema = [
	{
		type: 'object',
		properties: {
			sortCharacterClasses: {
				type: 'boolean',
				default: true
			}
		}
	}
];

module.exports = {
	create,
	meta: {
		type: 'suggestion',
		docs: {
			description: 'Improve regexes by making them shorter, consistent, and safer.',
			url: getDocumentationUrl(__filename)
		},
		fixable: 'code',
		schema,
		messages
	}
};
```

I'll look into it.

fisker:
Fixes #110

### DIFF
--- a/lib/rules/require-meta-schema.js
+++ b/lib/rules/require-meta-schema.js
@@ -60,9 +60,13 @@ module.exports = {
           });
           return;
         }
+
         let { value } = schemaNode;
         if (value.type === 'Identifier') {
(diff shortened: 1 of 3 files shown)


aladdin-add (approved):
LGTM, thanks!

PR #112, opened 2021-04-16
Refactor: `utils.getRuleInfo`

fisker:
Make it accept object contains `ast` and optional `scopeManager`, so we can simply just pass `sourceCode` to it. I believe this will fix some reference searching.

### DIFF
--- a/lib/rules/meta-property-ordering.js
+++ b/lib/rules/meta-property-ordering.js
@@ -27,7 +27,7 @@ module.exports = {
 
   create (context) {
     const sourceCode = context.getSourceCode();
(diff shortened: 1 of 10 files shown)


aladdin-add:
just to confirm: `sourceCode` contains more than `ast` and `scopeManager`, but other properties will not be used?

I feel a little hesitate to pass the unused properties - it could be confusing. :)

fisker:
> just to confirm: sourceCode contains more than ast and scopeManager, but other properties will not be used?

Yes. But accessing `sourceCode` is cheap, almost every rule need access to it.

aladdin-add (approved):

PR #113, opened 2021-05-08
Improve detection of static `description` strings and ignore non-static descriptions in `require-meta-docs-description` rule

### LINKED ISSUE #97: require-meta-docs-description does not support concatenated strings
For example:
```js
description: 'disallow ' + utils.jQueryGlobalLink( 'fx.interval' ) + '.'
```

When strings are concatenated, we can at least check leading string literals.

bmish:
Two changes:
* Drastically improves detection of the `description` string value when it can be determined statically, including handling when the description is stored in a constant or created through string concatenation
* If we cannot determine the string value of the `description` statically, then bail out and report no violations, so that users can dynamically generate the description if they'd like to

Fixes #97.

### DIFF
--- a/lib/rules/require-meta-docs-description.js
+++ b/lib/rules/require-meta-docs-description.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const { getStaticValue } = require('eslint-utils');
 const utils = require('../utils');
 
 // ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


aladdin-add (approved):
LGTM, thanks!

aladdin-add:
released in v3.0.3!

PR #121, opened 2021-06-14
Mention allowed `type` values in `require-meta-type` rule doc

bmish:
The list and explanations come from: https://eslint.org/docs/developer-guide/working-with-rules#rule-basics


aladdin-add (approved):
LGTM, thanks!

PR #123, opened 2021-06-14
Elaborate on `output` test assertion requirement and benefits in `consistent-output` rule doc

bmish:


aladdin-add (approved):
LGTM, thanks!

PR #127, opened 2021-06-14
Switch from `.eslintrc.yml` to `.eslintrc.js`

bmish:
No behavior change.

The JavaScript format of this file is a lot more common these days and it is easier to work with in that we can dynamically generate parts of it.

In particular, it will enable a follow-up to dynamically read the package version when enforcing rule URLs: https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-docs-url.md#version-specific-url


aladdin-add (commented):
can you also put them in 1 config? (the merging behavior will be removed in the new config system.)

bmish:
@aladdin-add good idea, updated.

aladdin-add (approved):
LGTM, thanks!

PR #128, opened 2021-06-14
Adopt `require-meta-docs-url` lint rule internally

bmish:
Now we can use our own lint rule [require-meta-docs-url](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-docs-url.md) to enforce and autofix the URL on all our lint rules. This means we no longer need to manually modify the lint rule object when exporting it.


package.json
```diff
@@ -9,7 +9,8 @@
     "lint": "eslint . --ignore-pattern \"!.*\"",
     "generate-readme-table": "node build/generate-readme-table.js",
     "generate-release": "node-release-script",
-    "test": "mocha tests --recursive"
+    "test": "mocha tests --recursive",
+    "version": "eslint --fix . && git add ."
```
aladdin-add, line 13: uh, how did this update the rule URLs during publishing?

bmish: Our own rule doc explains it :) https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-docs-url.md#version-specific-url

aladdin-add: didn't noticed the rule was fixable. makes sense to me!

aladdin-add: as we are using npm scripts `generate-release` to update the version. not sure it could run `npm version`.

will test it on my local later.

aladdin-add: I had tested on my local, it did not work well - seems `npm version` is not called in https://www.npmjs.com/package/@not-an-aardvark/node-release-script.

what do you think just add a function like `getRuleDocUrl()` in the utils?

bmish: We could use a util function `getRuleDocUrl() { /* return URL with current package version */ }` but that doesn't seem like much of an improvement over the status quo to me. It wouldn't be autofixable so users would have to know to call this function.

It looks like `@not-an-aardvark/node-release-script` has minimal usage and isn't maintained. I use the popular [release-it](https://github.com/release-it/release-it) in my projects for the same purpose and it correctly handles the standard `version` script.

aladdin-add: 👍 I like it!
I think the reason using [@not-an-aardvark/node-release-script](https://www.npmjs.com/package/@not-an-aardvark/node-release-script) is to keep the same commit format with eslint repo. But the eslint team has a plan to [switch to the conventional commits](https://github.com/eslint/eslint/issues/14231). so, it seems no longer an issue. cc @not-an-aardvark

bmish: Cool, should I help with switching to `release-it` or do you want to handle that? We can merge this PR in after that's done.

aladdin-add: sounds good!

PR #129, opened 2021-06-15
Improve detection of static arguments of `context.report()` in several rules

bmish:
Several of our rules check the arguments of `context.report()`. This PR makes some improvements:

1. When inferring the arguments passed to `context.report()` in `utils.getReportInfo()`, take into account the static value of the second argument when available.
2. Update the rules using `utils.getReportInfo()` to also consider the static value or variable declaration of the message argument when necessary.

I noticed these issues because in a number of my plugins, we have stored rule error messages in variables like `const MESSAGE = 'do A instead of B';`, and this prevented many of the eslint-plugin-eslint-plugin rules from operating correctly.

Example of how this improvement enables the `no-deprecated-report-api` rule to correctly autofix this code (previously no autofix):

```js
const MESSAGE = 'do A instead of B';
module.exports = {
    create(context) {
		// Before autofix
    	context.report(theNode, MESSAGE);

		// After autofix
		context.report({node: theNode, message: MESSAGE});
	}
};
```

### DIFF
--- a/lib/rules/no-deprecated-report-api.js
+++ b/lib/rules/no-deprecated-report-api.js
@@ -48,7 +48,7 @@ module.exports = {
             fix (fixer) {
               const openingParen = sourceCode.getTokenBefore(node.arguments[0]);
               const closingParen = sourceCode.getLastToken(node);
-              const reportInfo = utils.getReportInfo(node.arguments);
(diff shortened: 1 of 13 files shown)


aladdin-add (approved):
LGTM, thanks!

PR #130, opened 2021-06-15
Add `markdownlint` for doc formatting

bmish:
Add popular tool for enforcing best markdown practices: https://github.com/igorshubovych/markdownlint-cli

Autofixed most violations.

### DIFF
--- a/docs/rules/test-case-shorthand-strings.md
+++ b/docs/rules/test-case-shorthand-strings.md
@@ -169,8 +169,7 @@ ruleTester.run('example-rule', rule, {
 });
 ```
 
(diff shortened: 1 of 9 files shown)


docs/rules/test-case-shorthand-strings.md
```diff
@@ -169,6 +169,7 @@ ruleTester.run('example-rule', rule, {
 });
 ```
 
+<!-- markdownlint-disable-next-line no-duplicate-header -->
 #### `never`
```
aladdin-add, line 173: seems this line should be `consistent-as-needed`. :)

bmish: Oh, nice it caught a bug!

PR #135, opened 2021-06-17
Add eslint-plugin-mocha

bmish:
Autofixes this rule: https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-mocha-arrows.md

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,6 +40,13 @@ module.exports = {
     {
       files: ['tests/**/*.js'],
       env: { mocha: true },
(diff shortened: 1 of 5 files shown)


aladdin-add:
I'm not sure we want to use the plugin - mocha is not widely used in it at all, as rules (the most part of the it) were tested using eslint's rule tester. /cc @not-an-aardvark

bmish:
Sounds good! Yeah I agree it's not of great value here.

PR #136, opened 2021-06-18
Add eslint-plugin-eslint-comments

bmish:
https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,6 +5,7 @@ module.exports = {
   plugins: ['node'],
   extends: [
     'not-an-aardvark/node',
+    'plugin:eslint-comments/recommended',
     'plugin:node/recommended',
     'plugin:unicorn/recommended',
   ],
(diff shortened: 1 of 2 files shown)


aladdin-add:
I do see the value of the plugin, but given we [did not used too many inline configs](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/search?q=eslint-disable), I don't feel strong to add a plugin for this. :)

bmish:
I do see 3 places in the codebase where we use `// eslint-disable-line` today. But mostly it's not existing code that I'm worried about, instead the primary benefit would be seen in future PRs that contributors open that may not be using the best eslint comment practices.

If you still don't think it's worthwhile, feel free to close this.

aladdin-add:
the point is: it is not recommended to use too many inline configs.  If an inline config is widely used, we better consider if it should be enabled/disabled in the whole repo.

PR #137, opened 2021-06-18
Add rule documentation consistency tests

bmish:
Adds tests to ensure that our rule docs have notices to indicate whether they are fixable and/or in the recommended config. These notices match the format used by ESLint itself. Example: https://eslint.org/docs/rules/no-extra-semi

Also adds tests for various other aspects of rule docs and that each rule has a documentation file and test file.


aladdin-add (approved):
looks good! 👍

PR #138, opened 2021-06-19
Flag a violation when rule options are used but an empty schema is specified in `require-meta-schema` rule

bmish:
Flags a violation if an empty schema is specified but rule options are used.

This new check is similar to how:
* `require-meta-fixable` checks that `fixable: true` is specified when an autofixer is found 
* `require-meta-has-suggestions` checks that `hasSuggestions: true` is specified when suggestions are provided

Note: this is *not* a breaking change, since ESLint itself already prevents rules from using options when an empty rule schema is specified, so it's not possible that this new violation could be flagged on existing rules.

### DIFF
--- a/docs/rules/require-meta-schema.md
+++ b/docs/rules/require-meta-schema.md
@@ -20,6 +20,13 @@ module.exports = {
   meta: { schema: null },
   create (context) {/* ... */},
 };
+
+module.exports = {
+  meta: { schema: [] },
+  create (context) {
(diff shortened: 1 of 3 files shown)


aladdin-add:
yes! per eslint's versioning policy(https://github.com/eslint/eslint#semantic-versioning-policy), it can be released in a minor.

aladdin-add (approved):
LGTM, thanks! 💯

PR #139, opened 2021-06-19
Enable `meta-property-ordering` rule internally

bmish:
This was autofixed.

We should use our own rules internally as an additional means of testing/dogfooding them in real-world usage.


aladdin-add (approved):
LGTM, thanks!

PR #141, opened 2021-06-21
Add sort-package-json

bmish:
https://www.npmjs.com/package/sort-package-json

Adds this dev dependency with two new NPM scripts to run/fix the package.json field sorting checks:

```
"lint:package-json": "sort-package-json --check",
"lint:package-json:fix": "sort-package-json",
```

If you don't find enough value to this addition, feel free to close.


aladdin-add:
yes, I don't think it's worthy to introduce a dependency. or just run `npx` to reorder the fields?

PR #143, opened 2021-06-23
Improve detection of fix functions that never return a fix in `fixer-return` rule

bmish:
### Goal of the [fixer-return](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/fixer-return.md) rule: 

Report a violation on fix functions that never return a fix.

### Behavior changes:

1. Catch fix functions that have a return statement but never actually return a fix (i.e. fix functions that only ever return undefined or similar values that are obviously not fixes).
2. Allow fix functions to have some code paths that do not return a fix (it's allowed for a fix function to bail out / return early and avoid fixing in some unsafe situations).

I would consider these both bug fixes to reduce false positives and false negatives.

This PR also increases the test code coverage for this rule up to 100%.

### Test cases that were invalid before and are now valid:

```
module.exports = {
  create (context) {
    context.report({
      fix (fixer) {
        if (foo) {
          return; // no autofix in this case
        }
        return fixer.foo();
      },
    });
  },
};
```

### Test cases that were valid before and are now invalid:

```
module.exports = {
  create (context) {
    context.report({
      fix (fixer) {
        return undefined;
      },
    });
  },
};
```

```
module.exports = {
  create (context) {
    context.report({
      fix (fixer) {
        return null;
      },
    });
  },
};
```

```
module.exports = {
  create (context) {
    context.report({
      fix (fixer) {
        return;
      },
    });
  },
};
```

```
module.exports = {
  create (context) {
    context.report({
      fix (fixer) {
        return [];
      },
    });
  },
};
```

```
module.exports = {
  create (context) {
    context.report({
      fix (fixer) {
        const returnValue = undefined;
        return returnValue;
      },
    });
  },
};
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Then configure the rules you want to use under the rules section.
 Name | ✔️ | 🛠 | Description
 ----- | ----- | ----- | -----
 [consistent-output](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/consistent-output.md) |  |  | enforce consistent use of output assertions in rule tests
-[fixer-return](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/fixer-return.md) | ✔️ |  | require fixer function to always return a value.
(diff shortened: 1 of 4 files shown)


aladdin-add (approved):
Looks great, thanks! 💯

PR #144, opened 2021-06-23
Update `fixer-return` rule to handle arrow function expressions

bmish:
Also improves test coverage including updating the tests to ensure each violation is reported at the correct location (since the location differs depending on the function type).

The lack of coverage for arrow functions was a bug / omission in the rule.


aladdin-add (approved):
LGTM, thanks!

PR #148, opened 2021-06-25
Chore: Update `eslint-utils` to v3

bmish:
This is technically a breaking change because we still support Node 13 but eslint-utils v3 does not.

https://github.com/mysticatea/eslint-utils/releases/tag/v3.0.0

Planned for v4 release (https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/120).


aladdin-add:
this is a non-breaking change after https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/151 get merged.

PR #149, opened 2021-06-26
Add `node_modules` to `.eslintignore`

bmish:
I saw something in this folder getting linted locally.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+node_modules/


aladdin-add:
I don't think it's needed, as eslint ignores node_modules/ by default.

> In addition to any patterns in the .eslintignore file, ESLint always follows a couple of implicit ignore rules even if the --no-ignore flag is passed. The implicit rules are as follows:
> 
> * node_modules/ is ignored.
> * dot-files (except for .eslintrc.*), as well as dot-folders and their contents, are ignored.

https://eslint.org/docs/user-guide/configuring/ignoring-code#the-eslintignore-file

bmish:
@aladdin-add that's what I thought. I figured out the actual issue though and opened up a replacement fix: #154

Issue #120, opened 2021-06-13
V4 Planned Changes


bmish:
Apologies if these plans are already tracked somewhere, but I wanted to put together a list of the changes I'm hoping to see in the next major version. What do you think?

* Upgrade to [eslint-utils v3](https://github.com/mysticatea/eslint-utils/releases/tag/v3.0.0)
* New Node engine requirement of `"node": "12.* || 14.* || >= 16.*"` (Node 10 end-of-life 2021-04 and Node 15 end-of-life 2021-06-01)
* Switch to [pure ES modules](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm)? I haven't investigated how/if this would work yet
* Update rule options:
  * Update [consistent-output](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/consistent-output.md) rule option default to `always` since it's always best practice to assert the `output`:
    * [As of ESLint 7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#additional-validation-added-to-the-ruletester-class), the `output` property is required for test cases that provide an autofix
    * And even when test cases don't autofix, it's still best practice to assert that there's no autofix with `output: null`
  * Add `meta.hasSuggestions` property to [meta-property-ordering](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/meta-property-ordering.md) rule (https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/119)
* Add new `recommended` rules:
  * I would actually add ALL existing rules. I've used all of this plugin's rules successfully in a number of my eslint plugins, and I find them pretty much all useful and agreeable. Only about 1/3 of existing rules are enabled as recommended right now, and I see this as a missed opportunity, since most people only bother to enable the recommended rules. Enabling more rules means more exposure/battle-hardening of the rules and more benefit to users.

bradzacher:
> I would actually add ALL existing rules. I've used all of this plugin's rules successfully in a number of my eslint plugins, and I find them pretty much all useful and agreeable.

The problem you've got is "pretty much".
The recommended set needs to be unopinionated.
The last thing you want is for someone to disable a rule that's recommended.

Examples of rules that are opinionated: require-meta-type, report-message-format, test-case-property-ordering, test-case-shorthand-strings.

[...]

PR #150, opened 2021-06-26
Breaking: Rename `consistent-output` rule to `require-test-output`

bmish:
Changes the default rule behavior to always require a test `output` assertion. The old default  behavior of `consistent` (only require `output` assertions if some tests have it) is still provided as an available option.

* [As of ESLint 7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#additional-validation-added-to-the-ruletester-class), the `output` property is required for test cases that provide an autofix
* And even when test cases don't autofix, it's still best practice to assert that there's no autofix with `output: null`

Part of v4 release (#120).


bmish:
I've reconsidered this change and I don't think it's necessary anymore since ESLint 7 now enforces that the `output` is asserted when a test case has an autofix. That should be enough to ensure autofixing is tested, while not requiring that every test case has an output assertion for non-autofixable rules.

PR #151, opened 2021-06-26
Breaking: Drop support for Node 10, 13, 15

bmish:
We now support Node 12, 14, 16+.

* Node 10 end-of-life 2021-04-30
* Node 13 end-of-life 2020-06-01
* Node 15 end-of-life 2021-06-01

https://nodejs.org/en/about/releases/
https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md
https://en.wikipedia.org/wiki/Node.js#Releases

- [x] Should we raise these versions even higher? [ESM for example](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm) requires `"^12.20.0 || ^14.13.1 || >=16.0.0"` (update: not now, see comments)

Part of the v4 release (https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/120).


package.json
```diff
@@ -61,6 +61,6 @@
     "eslint": "^7.0.0"
   },
   "engines": {
-    "node": "^10.12.0 || >=12.0.0"
+    "node": "12.* || 14.* || >= 16.*"
```
MichaelDeBoey, line 64: I think we should bring them in line with ESLint 8.0.0

```suggestion
    "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
```

MichaelDeBoey (changes requested):

package.json
bmish: ESLint raised the requirement to these higher versions mostly so they could use ESM code/dependencies.

I'm open to using these versions for eslint-plugin-eslint-plugin, but I'm not aware of a specific need/benefit from us doing so at this time.

If we converted eslint-plugin-eslint-plugin to ESM soon, which could be a fairly-large task if it's even practical right now, that would be a good reason to raise the versions. But we could also do this in a subsequent major version if we don't get to it now.

aladdin-add: agreed. and the plugin supports multi eslint versions - does not have to be in line with newest eslint.

bmish: Yeah, I think I would rather we allow all Node 12/14/16 users to use this linting plugin, until we eventually convert it to ESM. So I'll leave the code as-is.

PR #155, opened 2021-07-06
Fix: Fix false negatives and reporting location in `require-meta-type` rule

bmish:
* Fixes false negatives
* Improves the reporting location

### DIFF
--- a/tests/lib/rules/require-meta-type.js
+++ b/tests/lib/rules/require-meta-type.js
@@ -37,6 +37,25 @@ ruleTester.run('require-meta-type', rule, {
         create(context) {}
       };
     `,
+    `
+      const type = 'problem';
+      module.exports = {
+        meta: { type },
+        create(context) {}
+      };
+    `,
+    `
+      module.exports = {
+        meta: { type: getType() },
+        create(context) {}
+      };
+    `,
+    `
+      module.exports = {
+        meta: { type: FOO },
+        create(context) {}
+      };
+    `,
     `module.exports = {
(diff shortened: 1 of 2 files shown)


tests/lib/rules/require-meta-type.js
```diff
@@ -99,7 +118,35 @@ ruleTester.run('require-meta-type', rule, {
...
+    },
+    {
+      code: `
+        const type = 'invalid-type';
+        module.exports = {
+          meta: { type },
+          create(context) {}
+        };
+      `,
+      errors: [{ messageId: 'unexpected', type: 'Identifier' }],
```
bmish, line 131: This test case was valid before (false negative).

tests/lib/rules/require-meta-type.js
```diff
@@ -99,7 +118,35 @@ ruleTester.run('require-meta-type', rule, {
...
+      errors: [{ messageId: 'unexpected', type: 'Literal' }],
+    },
+    {
+      code: `
+        module.exports = {
+          meta: { type: undefined },
+          create(context) {}
+        };
+      `,
+      errors: [{ messageId: 'unexpected', type: 'Identifier' }],
```
bmish, line 149: This test case was valid before (false negative).

aladdin-add (approved):
LGTM, thanks! 💯

bmish:
@aladdin-add thanks! Several other rules have the same problems so I'll be submitting more PRs in the next few days.

PR #158, opened 2021-07-09
Fix: Fix false positives/negatives in `require-meta-fixable` rule

bmish:
False positives fixed:

* When `meta.fixable` uses a variable

False negatives fixed:

* When there's a fixer but `meta.fixable` = null
* When there's a fixer but `meta.fixable` = undefined

Also improves reporting location.

I'll be opening a few more PRs after this to make similar fixes to other rules.

### DIFF
--- a/docs/rules/require-meta-fixable.md
+++ b/docs/rules/require-meta-fixable.md
@@ -2,11 +2,11 @@
 
 ✔️ The `"extends": "plugin:eslint-plugin/recommended"` property in a configuration file enables this rule.
 
-A fixable ESLint rule must have a valid `meta.fixable` property. A rule reports a problem with a `fix()` function but does not export a `meta.fixable` property is likely to cause an unexpected error.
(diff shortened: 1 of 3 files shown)


docs/rules/require-meta-fixable.md
```diff
@@ -44,20 +44,6 @@ module.exports = {
...
-  context.report({
-    node,
-    message: 'foo',
-    fix (fixer) {
-      return fixer.remove(node);
-    },
-  });
-} };
-```
-
```
aladdin-add, line 60: why removing the example? it seems a good example to have.

bmish: This example shows the rule without any `meta` object. I would this to be a rare edge case, and I would argue that including too many unimportant edge cases can distract from the primary examples and make the documentation too long and difficult to understand. The key point we're trying to get across is that `fixable` being missing is in invalid case, and that's already covered by the test case above this one.

If you still prefer, I can add it back, but just wanted to make my case against it.

aladdin-add (approved):
LGTM, thanks!

PR #162, opened 2021-07-10
Fix: Improve detection of static `url` strings in `require-meta-docs-url` rule

bmish:
* Add static value detection for `url` variables
* Ignore when we can't statically determine the value of a given `url`, so people can use their own `getUrl()` function if they want
* Improve reporting location and tests for this
* Use `messageId`
* Add tests

### DIFF
--- a/lib/rules/require-meta-docs-url.js
+++ b/lib/rules/require-meta-docs-url.js
@@ -55,7 +55,7 @@ module.exports = {
 
     /**
      * Check whether a given URL is the expected URL.
-     * @param {String} url The URL to check.
+     * @param {string} url The URL to check.
      * @returns {boolean} `true` if the node is the expected URL.
      */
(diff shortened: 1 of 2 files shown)


bmish:
Thanks for reviewing the PRs @aladdin-add! I have another one to open after this one.

lib/rules/require-meta-docs-url.js
```diff
@@ -48,24 +54,22 @@ module.exports = {
       : options.pattern.replace(/{{\s*name\s*}}/g, ruleName);
 
     /**
-     * Check whether a given node is the expected URL.
-     * @param {Node} node The node of property value to check.
+     * Check whether a given URL is the expected URL.
+     * @param {String} url The URL to check.
```
aladdin-add, line 58:
```suggestion
     * @param {string} url The URL to check.
```

PR #164, opened 2021-07-11
Fix: Ensure `require-meta-*` rules test `null` / `undefined` property values

bmish:
We need to ensure that the rules correctly handle `null` / `undefined` as the value of the meta properties since these values are commonly used as placeholder values.

Fixes one bug and adds test coverage.


aladdin-add (approved):
💯

aladdin-add:
a conflict needs to be resolved.

bmish:
Fixed.

PR #165, opened 2021-07-11
New: Add `catchNoFixerButFixableProperty` option (default false) to catch non-fixable rules that enable the fixable property in `require-meta-fixable` rule

bmish:
Added behind an option due to risk of false positives.

Open to better names for the option.

### DIFF
--- a/tests/lib/rules/require-meta-fixable.js
+++ b/tests/lib/rules/require-meta-fixable.js
@@ -157,6 +157,20 @@ ruleTester.run('require-meta-fixable', rule, {
       `,
       options: [{ catchNoFixerButFixableProperty: false }],
     },
+    // catchNoFixerButFixableProperty = true
+    {
+      code: `
+        module.exports = {
+          meta: { fixable: 'code' },
+          create(context) {
+            foo
+              ? context.report({ node, message })
+              : context.report({ node, message, fix });
+          }
+        };
+      `,
+      options: [{ catchNoFixerButFixableProperty: true }],
+    },
   ],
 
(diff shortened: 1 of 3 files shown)


bmish:
I think the solution to this is to put it behind an option (default off) since it may be too aggressive for some people. Will update.

tests/lib/rules/require-meta-fixable.js
```diff
@@ -207,5 +232,27 @@ ruleTester.run('require-meta-fixable', rule, {
...
+    {
+      code: `
+        module.exports = {
+          meta: { fixable: 'whitespace' },
+          create(context) { context.report({node, message}); }
+        };
+      `,
+      options: [{ catchNoFixerButFixableProperty: true }],
+      errors: [{ messageId: 'noFixerButFixableValue', type: 'Literal' }],
+    },
```
aladdin-add, line 256: can you add some tests that sometimes do not have a fixer - it seems a quite common case?
e.g.
```js
// valid
module.exports = {
  meta: { fixable: "whitespace" },
  create(context) {
    foo
      ? context.report({ node, message, fix })
      : ontext.report({ node, message });
  },
};

// invalid
module.exports = {
  meta: { fixable: null },
  create(context) {
    foo
      ? context.report({ node, message, fix })
      : ontext.report({ node, message });
  },
};

```

bmish: Sure, added tests.

aladdin-add (approved):
LGTM, thanks! 💯

PR #170, opened 2021-07-24
New: Add new rule `prefer-message-ids`

bmish:
* Adds new rule (see rule doc)
* Updates other rules to comply with it


aladdin-add (approved):
it can replace the eslint's internal rule: https://github.com/eslint/eslint/blob/master/tools/internal-rules/consistent-meta-messages.js

LGTM, thanks!

bmish:
@aladdin-add yeah, that was actually the inspiration for it!

PR #171, opened 2021-07-27
Fix: Fix false positive with empty array variable in `require-meta-has-suggestions` rule

bmish:
If the suggestions array is initialized as an empty array variable, we need to assume that someone may have pushed suggestion objects to it, so we should consider that a usage of suggestions.

Previously, this rule incorrectly assumed that using `const suggest = []; ... context.report({suggest,...});` meant the rule was not using suggestions.

Real world example here: https://github.com/eslint/eslint/blob/master/lib/rules/no-nonoctal-decimal-escape.js#L91


aladdin-add (approved):

aladdin-add:
released in eslint-plugin-eslint-plugin@3.5.1

bmish:
Thanks!

PR #172, opened 2021-07-29
Fix: support eslint v6

aladdin-add:
refs: https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/95#issuecomment-888410721


aladdin-add:
As of now, the package does not support multi eslint versions. I have tested eslint v6 on my local, and it's working fine.

PR #173, opened 2021-07-29
Fix: False negative in `prefer-message-ids` rule

bmish:
In some rules, `context` is passed to functions that eventually call `context.report()`, like [`no-duplicate-imports`](https://github.com/eslint/eslint/blob/8be8a36010145dfcd31cbdd4f781a91989e3b1bd/lib/rules/no-duplicate-imports.js#L176) or [`padding-lines-between-statements`](https://github.com/eslint/eslint/blob/8be8a36010145dfcd31cbdd4f781a91989e3b1bd/lib/rules/padding-line-between-statements.js#L244), and that will prevent this existing rule from flagging such instances of `context.report({ message: ... });`.

So in addition to flagging that, we should also flag when rules are simply missing `meta.messages`, as all rules using message IDs should have at least one violation message defined in that object.

This will unblock this PR: https://github.com/eslint/eslint/pull/14841#issuecomment-888645170

CC: @mdjermanovic

### DIFF
--- a/lib/rules/prefer-message-ids.js
+++ b/lib/rules/prefer-message-ids.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const utils = require('../utils');
+const { getStaticValue } = require('eslint-utils');
 
 // ------------------------------------------------------------------------------
 // Rule Definition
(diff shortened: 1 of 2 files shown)


aladdin-add (approved):
LGTM, thanks!

aladdin-add:
released in v3.5.3. 🍺

PR #177, opened 2021-09-21
Breaking: Support ESM rules

### LINKED ISSUE #157: support esm
* some plugins were written in esm. e.g.  [eslint-plugin-react-hooks](https://github.com/facebook/react/blob/c62986cfd8cbf994348407c8ece06b04ec1c86f4/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js)
* plugins authors can write native esm once the new config system landed. refs: https://github.com/eslint/eslint/issues/13481 https://github.com/eslint/rfcs/pull/9

bmish:
Fixes #157. I recommend we add this to the upcoming v4 release (#120).

This change updates most rules (rules using our util `getRuleInfo` to read rule `create`/`meta` data) to support rules defined using the ES modules syntax as follows:

```js
export default {
  meta: {},
  create(context) {}
};
```

TODO:

- [x] Add tests for all affected rules.
- [x] ~~Support variables.~~ We've never supported this so I don't think it's essential now. We can follow-up to try to add it later if we want.
  ```
  const rule = { meta: {}, create(context) {} }; 
  export default rule;
  ```
- [x] Support `create` function as the default export.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -83,8 +83,93 @@ function isRuleTesterConstruction (node) {
   );
 }
 
-module.exports = {
+const INTERESTING_RULE_KEYS = new Set(['create', 'meta']);
 
+/**
+ * Helper for `getRuleInfo`. Handles ESM rules.
+ */
+function getRuleExportsESM (ast) {
+  return ast.body
+    .filter(statement => statement.type === 'ExportDefaultDeclaration')
+    .map(statement => statement.declaration)
+    // eslint-disable-next-line unicorn/prefer-object-from-entries
+    .reduce((currentExports, node) => {
+      if (node.type === 'ObjectExpression') {
+        // eslint-disable-next-line unicorn/prefer-object-from-entries
+        return node.properties.reduce((parsedProps, prop) => {
+          const keyValue = module.exports.getKeyName(prop);
+          if (INTERESTING_RULE_KEYS.has(keyValue)) {
+            parsedProps[keyValue] = prop.value;
+          }
+          return parsedProps;
+        }, {});
+      } else if (isNormalFunctionExpression(node)) {
+        return { create: node, meta: null, isNewStyle: false };
+      }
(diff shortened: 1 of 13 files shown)


lib/utils.js
```diff
@@ -150,6 +171,8 @@ module.exports = {
         return currentExports;
       }, {});
 
+    const exportNodes = exportNodesESM || exportNodesCJS;
```
aladdin-add, line 174: when `ast.sourceType === "module"`, it seems no need to calc `exportNodesCJS`?

what do you think checking its value like: `exportNodes = ast.sourceType === "module" ? calcEsmExport() : calcCjsExport();`?

bmish: Will update.

bmish: Updated.

tests/lib/utils.js
```diff
@@ -110,11 +125,35 @@ describe('utils', () => {
...
+          create: { type: 'ArrowFunctionExpression' },
+          meta: null,
+          isNewStyle: false,
+        },
       };
 
       Object.keys(CASES).forEach(ruleSource => {
         it(ruleSource, () => {
-          const ast = espree.parse(ruleSource, { ecmaVersion: 6, range: true });
+          const ast = espree.parse(ruleSource, { ecmaVersion: 6, range: true, sourceType: ruleSource.includes('export default') ? 'module' : 'script' });
```
aladdin-add, line 156: it seems very not reliable to just check 'export default'.  imagine this comment: `// export default xxxx`.

can we get parser options from eslint here?

bmish: I tweaked it a bit to check that the test case starts with `export default` which should be safer. I know it's not great, but this is just for a single test.

aladdin-add (approved):

aladdin-add:
we need  to resolve the conflicts. :)

PR #178, opened 2021-09-23
Fix: Remove erroneous schema from `require-meta-schema` rule

bmish:
Ironically, this rule had the wrong schema. This rule has no options and should not have an empty schema. The schema used here was supposed to be an example in the rule doc file only.


bmish:
@aladdin-add can you merge this? It's blocking my next change to add an option to this rule (#179).

aladdin-add (approved):

PR #180, opened 2021-09-23
New: Add `requireSchemaPropertyWhenOptionless` option to `require-meta-schema` rule

### LINKED ISSUE #179: `require-meta-schema`: Add `requireSchemaPropertyWhenOptionless` option
Add `requireSchemaPropertyWhenOptionless` option to `require-meta-schema` rule (default `true` initially to match current behavior, could be changed in major release).

Why?

* Some people don't want to bother including `schema: []` when their rules have no options ([example](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/120#issuecomment-861155550))
* `schema: []` might not be necessary in the future if the ESLint default behavior changes (see below)

> I have an update to share regarding the [`require-meta-schema`](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-schema.md) rule. I have recommended to the ESLint team that [schemas should be required for rules with options](https://github.com/eslint/eslint/issues/14709) and they are interested in this idea so I'll be opening an RFC for it targeting ESLint 9.
>
> With this change, rules without options will no longer need to specify `schema: []`, as that will be the default behavior. That means we can update the `require-meta-schema` rule to only require the `schema` property for rules that actually have options (which is the same as how [`require-meta-fixable`](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-fixable.md) only requires the `fixable` property for rules with fixers).
>
> As a result of this change, we should be able to enable the `require-meta-schema` rule as `recommended` in eslint-plugin-eslint-plugin v4.

_Originally posted by @bmish in https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/120#issuecomment-877337139_

bmish:
Fixes #179.

### DIFF
--- a/lib/rules/require-meta-schema.js
+++ b/lib/rules/require-meta-schema.js
@@ -16,7 +16,18 @@ module.exports = {
       recommended: false, // TODO: enable it in a major release.
     },
     fixable: 'code',
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          requireSchemaPropertyWhenOptionless: {
+            type: 'boolean',
+            default: true,
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
     messages: {
       foundOptionsUsage: '`meta.schema` has no schema defined but rule has options.',
       missing: '`meta.schema` is required (use [] if rule has no schema).',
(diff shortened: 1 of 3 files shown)


lib/rules/require-meta-schema.js
```diff
@@ -33,27 +44,33 @@ module.exports = {
...
-            fix (fixer) {
-              return utils.insertProperty(fixer, metaNode, 'schema: []', sourceCode);
-            },
-          });
+          if (requireSchemaPropertyWhenOptionless) {
+            context.report({
+              node: metaNode,
+              messageId: 'missing',
+              fix (fixer) {
+                return utils.insertProperty(fixer, metaNode, 'schema: []', sourceCode);
```
aladdin-add, line 70: is it likely to change the runtime behavior?
before fixing: skip validating
after fixing: no options allowed.

or it can be a suggestion?

aladdin-add (approved):
LGTM~

Issue #183, opened 2021-09-24
`require-meta-schema`: Change autofix to suggestion


bmish:
> the autofix will introduce a breaking change for users when passing unknown options:
>
> before autofix: skip validating and lint working fine.
> after autofix: an error thrown from eslint.

Relevant: https://github.com/eslint/eslint#semantic-versioning-policy

_Originally posted by @aladdin-add in https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/182#issuecomment-926292475_

PR #184, opened 2021-09-24
Fix: Only autofix when no options present in `require-meta-schema` rule

### LINKED ISSUE #182: `require-meta-schema` should only autofix when no options present
> The goal is that the rule should only autofix to `schema: []` if no options are used.

_Originally posted by @bmish in https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/180#discussion_r715251282_

bmish:
Fixes #182.

After this is merged, I will follow-up to convert the autofix to a suggestion, as per #183.


aladdin-add (approved):
LGTM! 💯

PR #194, opened 2021-09-26
Breaking: Update `fixer-return` and `prefer-replace-text` rules to also apply to suggestion fixer functions

### LINKED ISSUE #190: Breaking: `fixer-return` should be updated to apply to suggestion fixer functions too
Right now, it only applies to autofix fixer functions. It should also apply to suggestion fixer functions.

https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions

### LINKED ISSUE #193: Breaking: `prefer-replace-text` should be updated to apply to suggestion fixer functions too
Right now, it only applies to autofix fixer functions. It should also apply to suggestion fixer functions.

https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions

bmish:
These rules should apply to both autofix and suggestion fixer functions.

Updating these two rules together since they share the same logic.

Fixes #193
Fixes #190

Part of v4 release (#120).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -424,4 +424,44 @@ module.exports = {
         ),
     ];
   },
+
+  /**
+   * Whether the provided node represents an autofixer function.
+   * @param {Node} node
+   * @param {Node[]} contextIdentifiers
+   * @returns {boolean}
+   */
+  isAutoFixerFunction (node, contextIdentifiers) {
+    const parent = node.parent;
+    return ['FunctionExpression', 'ArrowFunctionExpression'].includes(node.type) &&
+      parent.parent.type === 'ObjectExpression' &&
+      parent.parent.parent.type === 'CallExpression' &&
+      contextIdentifiers.has(parent.parent.parent.callee.object) &&
+      parent.parent.parent.callee.property.name === 'report' &&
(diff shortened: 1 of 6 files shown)


lib/utils.js
```diff
@@ -385,4 +385,44 @@ module.exports = {
...
+      parent.parent.parent.type === 'ArrayExpression' &&
+      parent.parent.parent.parent.type === 'Property' &&
+      parent.parent.parent.parent.key.type === 'Identifier' &&
+      parent.parent.parent.parent.key.name === 'suggest' &&
+      parent.parent.parent.parent.parent.type === 'ObjectExpression' &&
+      parent.parent.parent.parent.parent.parent.type === 'CallExpression' &&
+      contextIdentifiers.has(parent.parent.parent.parent.parent.parent.callee.object) &&
+      parent.parent.parent.parent.parent.parent.callee.property.name === 'report' &&
+      this.getReportInfo(parent.parent.parent.parent.parent.parent.arguments).suggest === parent.parent.parent;
+  },
```
aladdin-add, line 427: can you add a few tests for it, in `tests/lib/utils.js`?

bmish: Added tests.

aladdin-add (approved):
LGTM, thanks! 💯

PR #195, opened 2021-09-26
Breaking: Update `report-message-format` to also apply to suggestion messages

### LINKED ISSUE #192: Breaking: `report-message-format` should apply to suggestion messages
The [report-message-format](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/report-message-format.md) only applies to violation messages, but not suggestion messages.

As a breaking change, we should update it to apply to both violation and suggestion messages.
```
report-message-format: ['error', '^[A-Z].*\\.$']
```

As a follow-up (non-breaking change), we could add object-based options to the rule to allow separate violation/suggestion message requirements:

```
report-message-format: ['error', { error: '^[A-Z].*\\.$', suggestion: '^[A-Z].*\\.$' }]
```

bmish:
Fixes #192.

Part of v4 release (#120).

### DIFF
--- a/lib/rules/report-message-format.js
+++ b/lib/rules/report-message-format.js
@@ -83,12 +83,19 @@ module.exports = {
         ) {
           const reportInfo = utils.getReportInfo(node.arguments, context);
           const message = reportInfo && reportInfo.message;
+          const suggest = reportInfo && reportInfo.suggest;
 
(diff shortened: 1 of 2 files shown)


aladdin-add (approved):
LGTM! 💯

PR #196, opened 2021-09-26
Breaking: Update `no-missing-placeholders`, `no-unused-placeholders`, `prefer-message-ids`, `prefer-placeholders` rules to also apply to suggestion messages

### LINKED ISSUE #191: Breaking: Message rules should be updated to apply to suggestion messages
These should all be updated to consider both autofixer/suggestion messages:

- no-missing-placeholders
- no-unused-placeholders
- prefer-message-ids
- prefer-placeholders

bmish:
Rules should apply to both violation messages and suggestion messages.

Fixes #191.

Part of v4 release (#120).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -169,6 +169,17 @@ function getRuleExportsCJS (ast) {
     }, {});
 }
 
+/**
+ * Find the value of a property in an object by its property key name.
+ * @param {Object} obj
+ * @param {String} keyName
+ * @returns property value
+ */
+function findObjectPropertyValueByKeyName (obj, keyName) {
+  const property = obj.properties.find(prop => prop.key.type === 'Identifier' && prop.key.name === keyName);
+  return property ? property.value : undefined;
+}
+
 module.exports = {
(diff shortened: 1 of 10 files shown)


lib/utils.js
```diff
@@ -169,6 +169,17 @@ function getRuleExportsCJS (ast) {
...
+ * Find the value of a property in an object by its property key name.
+ * @param {Object} obj
+ * @param {String} keyName
+ * @returns property value
+ */
+function findObjectPropertyValueByKeyName (obj, keyName) {
+  const property = obj.properties.find(prop => prop.key.type === 'Identifier' && prop.key.name === keyName);
+  return property ? property.value : undefined;
+}
+
```
aladdin-add, line 182: same here, better to have a test for it. :)

bmish: Agreed, added a test for the new util function `collectReportViolationAndSuggestionData`.

aladdin-add (approved):
LGTM, thanks!

PR #197, opened 2021-09-27
Breaking: Add support for TypeScript rules

### LINKED ISSUE #176: TS support
Hello ✋

I was trying to use this library in a project, but as I didn't receive any reports (even though I knew there were violations there), I decided to look at the source code and I saw that some rules like `no-missing-placeholders` and `report-message-format` (those are the rules I was most interested in, but I believe there are much more) basically depend on `utils.getContextIdentifiers`, which looks for `module.exports` or `exports.create` (**please, correct if I'm wrong  — I have no familiarity with this repository code and this assumption comes from the little I could see by looking at the code + tests**), but with TS (more specifically `@typescript-eslint`), the rules are _usually_ created using the following structure:

```ts
// https://github.com/timdeschryver/eslint-plugin-ngrx
export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
  meta: {
    // ...
  }
})

// https://github.com/angular-eslint/angular-eslint
export default createESLintRule<Options, MessageIds>({
  meta: {
    // ...
  }
})

// https://github.com/typescript-eslint/typescript-eslint/issues
export default util.createRule<Options, MessageIds>({
  meta: {
    // ...
  }
})
```

So I wonder if the implementation could be changed to not depend specifically on `module.exports` or `exports.create` and to support these cases above as well. I really wanted to avoid redundancy and use some of the rules that already exist here instead of rewriting from scratch, but I understand if it's not within the scope of your project.

bmish:
Fixes #176.

Adds ability for us to lint rules written using several common TypeScript rule formats:

```js
// https://github.com/timdeschryver/eslint-plugin-ngrx
export default ESLintUtils.RuleCreator(docsUrl)<Options, MessageIds>({
  meta: {
    // ...
  }
})

// https://github.com/angular-eslint/angular-eslint
export default createESLintRule<Options, MessageIds>({
  meta: {
    // ...
  }
})

// https://github.com/typescript-eslint/typescript-eslint
export default util.createRule<Options, MessageIds>({
  meta: {
    // ...
  }
})
```

Note that we only look for this approximate structure (we don't actually verify util function names as those could be named anything).

Part of v4 release (#120).


aladdin-add:
To fully support ts, do we need to introduce ts-eslint/parser for testing?

bmish:
> To fully support ts, do we need to introduce ts-eslint/parser for testing?

I'll give that a try to be safe, although the parts of the rule AST that we care about have the same structure in JS/TS.

bmish:
This is ready.

PR #199, opened 2021-10-03
Breaking: Add `fixer-return` to `rules` preset

bmish:
All rules should have a category of either `Rules` or `Tests` since this is how those presets are generated.

Technically, this isn't a breaking change, since we're allowed to modify these presets any time, according to the README:

> The list of recommended rules will only change in a major release of this plugin. However, new non-recommended rules might be added in a minor release of this plugin. Therefore, the using the all, rules, and tests presets is not recommended for production use, because the addition of new rules in a minor release could break your build. 

Part of v4 release (#120).

### DIFF
--- a/lib/rules/fixer-return.js
+++ b/lib/rules/fixer-return.js
@@ -21,7 +21,7 @@ module.exports = {
     type: 'problem',
     docs: {
       description: 'require fixer functions to return a fix',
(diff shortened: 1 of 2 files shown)


aladdin-add (approved):
LGTM

PR #205, opened 2021-10-11
Docs: Move preset list into table in README

bmish:
Many other ESLint plugins use a table for listing presets/configs (similar to the rules table):
* https://github.com/ember-cli/eslint-plugin-ember#-configurations
* https://github.com/ember-template-lint/ember-template-lint#presets
* https://github.com/platinumazure/eslint-plugin-qunit#configurations

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,13 +112,3 @@ Or to apply linting only to the appropriate rule or test files:
     ]
 }
 ```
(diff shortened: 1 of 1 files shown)


aladdin-add:
ah, thanks! ❤️

PR #212, opened 2021-10-13
Breaking: Drop support for old minor versions of Node 12 and 14

bmish:
This matches [ESLint's Node version support](https://github.com/eslint/eslint/blob/c584a63e2d6d9c0a66e5c5a5d43bc8148c054f5d/package.json#L147). It sets us up for:

* Using pure ESM dependencies (which are quickly becoming commonplace)
* Converting this package to ESM (see Sindre's [post](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c))
* Using other new Node features like `node:` prefix imports

Especially because this plugin is a dev-dependency for consumers, we can be more aggressive/strict here.

Originally discussed here: https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/151#discussion_r693591610

Part of v4 release (#120).


aladdin-add:
not to against the change, just to clarify:

1. Sindre's packages require "node>=12". e.g. https://github.com/sindresorhus/is-unicode-supported/commit/30ba81f2c7d69e3f2b03df46eb9eb9b1c0d923c4#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R16
2. if no build process, converting to esm is a breaking change, even it requires "node >=12.22.0" (esm cannot be required).

bmish:
@aladdin-add Good points. Still, I think this PR is a worthwhile change.

PR #217, opened 2021-10-15
Docs: Improve CHANGELOG for v4

bmish:
I manually fixed a number of issues for the v4 release changelogs:

* Added summary to v4 entry so it can standalone from prerelease entries
* Alphabetized items, but moved "Breaking" and "Fixes" first since those are most important for consumers
* Removed duplicate entries from including issues (only PRs should be included) 
* Improved some entry titles for consistency

My goal is that we have a very easy-to-read changelog for v4 in a single place that draws attention to the most important changes first.

@aladdin-add can we also edit the text of https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/releases/tag/v4.0.0 to include the summary that I've created? That way, we can share the links to the v4 entry and people don't have to sort through the preleases to understand it.


aladdin-add:
the change was overwritten by [a new release 4.0.1](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/commit/51fa5bb6b033321de0565ccf3d6ddfdd03bddd48) v/_ \. we will need find a way to avoid it being changed.

PR #226, opened 2021-10-30
Ensure rule docs mention all rule options

bmish:
Add test to ensure:

1. Any rule with options has a `## Options` section in its rule doc.
2. Any named rule options (for example, the `pattern` option for the `require-meta-docs-description` rule) are mentioned in the rule doc.

Based on the same test I wrote here: https://github.com/ember-cli/eslint-plugin-ember/blob/4f26c0bc7e34400007e02939a9c6e183078a3e87/tests/rule-setup.js#L111

Eventually, I'd like to add a lint rule to enforce these "rule setup" tests for any eslint plugin.

### DIFF
--- a/docs/rules/require-meta-docs-url.md
+++ b/docs/rules/require-meta-docs-url.md
@@ -8,20 +8,6 @@ A rule can store the URL to its documentation page in `meta.docs.url`. This enab
 
 This rule aims to require ESLint rules to have a `meta.docs.url` property.
 
-This rule has an option.
-
-```json
-{
-  "eslint-plugin/require-meta-docs-url": ["error", {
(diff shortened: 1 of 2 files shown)


aladdin-add (approved):

aladdin-add:
> Eventually, I'd like to add a lint rule to enforce these "rule setup" tests for any eslint plugin.

eslint does not support `ext` in config files, so users have to use `eslint --ext=.js,.md`?

bmish:
> > Eventually, I'd like to add a lint rule to enforce these "rule setup" tests for any eslint plugin.
> 
> eslint does not support `ext` in config files, so users have to use `eslint --ext=.js,.md`?

More likely, the rule would be similar to https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/2, where a violation is reported in the rule JS file if the documentation file that we manually read in is missing or has missing sections.

PR #231, opened 2021-11-30
chore: use file:// to dogfood

aladdin-add:
the devdep eslint-plugin-self is no longer needed.

cc @bmish


aladdin-add:
the older eslint testing was failing - I'll take a look later.

bmish (approved):
Neat!

PR #233, opened 2021-12-07
fix: detect rules exported using a variable

bmish:
Updates our rules to detect rules that are defined in variables, which is a style I have seen used by some eslint plugins:

```js
const rule = { meta: {}, create(context) {} }; 
export default rule;
```

```js
const rule = { meta: {}, create(context) {} }; 
module.exports = rule;
```

Follow-up to previous work around detecting additional types of rules and variables used inside rules:
* https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/177
* https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/225

Note that this could be considered either a fix or a breaking change since it will result in additional rules detected by our rules in some plugins.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -83,9 +83,31 @@ function isFunctionRule (node) {
 }
 
 /**
+ * Check if the given node is a function call representing a known TypeScript rule creator format.
+ * @param {Node} node
+ * @returns {boolean}
+ */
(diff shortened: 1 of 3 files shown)


aladdin-add (approved):
LGTM, thanks!

PR #234, opened 2021-12-08
chore: add prettier

bmish:
Adds [prettier](https://prettier.io/) for automatic code formatting and runs it with eslint:
* package.json
* .eslintrc.js
* .prettierrc.js

Everything else is autofixed.


.eslintrc.js
```diff
@@ -2,10 +2,15 @@
 
 module.exports = {
   root: true,
+  parserOptions: {
+    ecmaVersion: 2021,
+    sourceType: 'script',
+  },
```
aladdin-add, line 8: why adding this - it should have been set by `plugin:node/recommended`. I've just put it to the last, to overwrite the config in other presets.

bmish: Oh, I see. [unicorn/recommended](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/configs/base.js) sets `sourceType: 'module'` so I had to set it back to `sourceType: 'script'` here.

This came up because I alphabetized the list in `extends`. I actually think it's better practice for us to be explicit about our `sourceType` instead of relying on the ordering of `extends`. We shouldn't have to worry about what's happening in `extends`.

But I can change it back if you prefer.

aladdin-add: I do like eslint-plugin-node provides a good default for most node.js apps, but surely, it does not hurt to add our own. :)

PR #235, opened 2021-12-13
fix: detect function-style rules exported using a variable

bmish:
Updates our rules to detect old function-style rules that are defined in variables:

```js
const rule = function(context) { return {} };
export default rule;
```

```js
const rule = function(context) { return {} };
module.exports = rule;
```

In particular, this is needed so that our [eslint-plugin/prefer-object-rule](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-object-rule.md) can detect and automatically convert these old function-style rules to the new object-style format.

Follow-up to:
* https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/233
* https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/177
* https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/225

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -141,6 +141,7 @@ function getRuleExportsESM(ast, scopeManager) {
             INTERESTING_RULE_KEYS
           );
         } else if (node.type === 'Identifier') {
+          // Rule could be stored in a variable before being exported.
           const possibleRule = findVariableValue(node, scopeManager);
           if (possibleRule) {
             if (possibleRule.type === 'ObjectExpression') {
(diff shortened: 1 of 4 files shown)


aladdin-add (approved):
LGTM, thanks! 💯

bmish:
Thanks for the review! Side node, I'd love to be considered for becoming an official collaborator on this project if you folks are open to it @not-an-aardvark. Looking forward to helping out a lot more.

aladdin-add:
ah, I was thinking the same. But I don't have the admin access too. /_ \

@not-an-aardvark can you give me the admin access, so I can add @bmish, and some more (e.g. rename default branch to main).

not-an-aardvark:
I added @bmish to the repository and renamed the default branch to `main`. Let me know if there are other things I should do.

I'm not sure it's possible for me to give admin access because this is a user-owned repository. I think we would need to transfer the repository to an organization in order to do that. I would be fine with transferring it to an organization if that makes things easier.

aladdin-add:
great, thanks! 👍 

nicholas has a plan to make an eslint-community org. this one sounds a good fit. I'll ping you when everything is ready. :)

bmish:
Thanks! Yeah, definitely in favor of this being part of the future eslint-community org (I suggested that here too: https://github.com/eslint/eslint/pull/14884#discussion_r683879412).

PR #236, opened 2021-12-14
chore: replaced chai to `assert/strict`

aladdin-add:


bmish (approved):
It looks like it was easy enough to convert to `node:assert` and it has what we need so I'm fine with this.

There are definitely some chai assertions that are really nice like [assert.include](https://www.chaijs.com/api/assert/#method_include) but it seems we don't need them yet.

aladdin-add:
yes, I'm closing it as they may be needed for further changes.

bmish:
Okay, I'm fine either way. We could always add it back later if we ever need it.

PR #238, opened 2021-12-19
Add dependency caching to CI workflow

ddzz:
Quick PR to add dependency caching to CI, which should speed up CI by at least a few seconds.

References:

* https://github.com/actions/setup-node#caching-packages-dependencies

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
+          cache: "npm"
       - run: npm install
       - run: npm test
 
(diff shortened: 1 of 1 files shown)


bmish (approved):
Thanks!

aladdin-add:
it is intentional to not caching - as stated in https://github.com/eslint/eslint#why-doesnt-eslint-lock-dependency-versions

bmish (changes requested):
Hmm, looks like this actually won't work / isn't appropriate for this repository because we don't have a lockfile, so we'll need to decline this.

```
Error: Dependencies lock file is not found in /home/runner/work/eslint-plugin-eslint-plugin/eslint-plugin-eslint-plugin. Supported file patterns: package-lock.json,yarn.lock
```

PR #245, opened 2022-03-28
Improve config example in README

### LINKED ISSUE #244: Plugin requires "parserOptions.sourceType" being configured to "script"
Rules in this repo have a dependency on eslint being configured with:

```
...
parserOptions: {
  sourceType: 'script',
}
```

This is makes sense, but I ran into unexpectedly as I created a plugin from scratch rather than using the yeoman generator (https://github.com/eslint/generator-eslint).
For context, using the generator `eslint-plugin-node/recommended` gets extended and this config sets the parserOption).

Can we update the documentation somewhere to make it clear that this is a configuration requirement?

I'm happy to create a PR to update `README.md` with this.

bmish:
* Provides single unified code sample in "Usage" section
* No more need to specify `plugins` since our configs already do that for the user
* Suggests using the `recommended` config in the "Usage" section instead of just manually enabling a recommended rule
* Demonstrates enabling an optional rule
* Explains when `sourceType` should be set to `script` vs. `module`

Fixes #244.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,22 +32,22 @@ npm install eslint-plugin-eslint-plugin --save-dev
 
 ## <a name='Usage'></a>Usage
 
-Add `eslint-plugin` to the plugins section of your `.eslintrc` configuration file. You can omit the `eslint-plugin-` prefix:
+Here's an example ESLint configuration that:
+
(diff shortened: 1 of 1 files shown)


tristanHessell:
@bmish This is the kind of thing I had in mind - much appreciated!

aladdin-add (approved):

PR #249, opened 2022-05-14
Fix: Detect nested test cases

### LINKED ISSUE #248: RuleTester detection seems like it could use improving
I've noticed the `RuleTester` detection logic seems to trip up on what I'd consider common or easy to support cases - the latest was that it didn't flag [this test](https://github.com/jest-community/eslint-plugin-jest/pull/1094/commits/a12fb7e745b7f0af5f70e6f7a05370981a488ce0) as duplicated because I've got the tests wrapped in a `describe`.

e.g.

```
describe('reference tests', () => {
  ruleTester.run('esm', rule, {
    valid: ['not flagged', 'not flagged'],
    invalid: [],
  });
});

ruleTester.run('esm', rule, {
  valid: ['not flagged', 'not flagged'],
  invalid: [],
});
```

I think there was at least another instance where it didn't seem to pick up alternative names, but it's been a while so can't remember the details.

I've not yet had the time to dig into this that deeply but wanted to open this issue in the meantime to get some discussion going - I'm happy to help with trying to improve the logic.

G-Rath:
I think this should cover most cases though it'd be easier if we were walking the AST tree like ESLint actually does, since really what we're wanting to do is check the body of all `BlockStatement` expressions but without walking we have to manually account for every possible place that could be held.

Still this already I think is a huge improvement - I've tried it on `eslint-plugin-jest` and it caught a few violations (including a couple I'd not noticed myself!)

Resolves #248

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -414,9 +414,22 @@ module.exports = {
 
     for (const statement of statements) {
       if (statement.type === 'VariableDeclaration') {
-        statement.declarations.forEach((declarator) => {
+        for (const declarator of statement.declarations) {
+          if (!declarator.init) {
+            continue;
+          }
+
+          const extracted = module.exports.extractFunctionBody(declarator.init);
+
+          runCalls.push(
+            ...module.exports.checkStatementsForTestInfo(
+              context,
+              extracted,
+              variableIdentifiers
+            )
+          );
+
           if (
-            declarator.init &&
             isRuleTesterConstruction(declarator.init) &&
             declarator.id.type === 'Identifier'
           ) {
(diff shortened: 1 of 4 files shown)


bmish (approved):
Nice!

aladdin-add (approved):
LGTM, thanks!

G-Rath:
No problem! Feel free to ping me if anyone reports any bugs or further improves as a result of this :)

aladdin-add:
released in [v4.2.0](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/releases/tag/v4.2.0) 🎉

PR #251, opened 2022-06-15
fix: handle properties behind spread syntax in `require-meta-*` rules

### LINKED ISSUE #250: `require-meta-type` don't support spread operator
When filling `meta` object using a spread operator instead of explicitly set the fields, the eslint validator fails. It should look for the content of the spread and see that the required fields are there, instead of forcing to set them explicitly.

bmish:
We're now able to take into account meta properties inside spreads in the `require-meta-*` rules.

1. Introduces new util function `evaluateObjectProperties` which gathers all properties from an object including those behind spreads.
2. Uses this new util function when checking for properties inside the meta object in the `require-meta-*` rules.

Fixes #250.

Example: this is now valid for the `require-meta-type` rule:

```
const extra = { type: 'problem' };
module.exports = {
    meta: { ...extra },
    create(context) {}
};
```

TODO:

- [x] Unit tests for `evaluateObjectProperties` util
- [x] Check for additional places we should evaluate spreads
- [x] Tests for autofixing behavior when properties are behind a spread

### DIFF
--- a/lib/rules/require-meta-docs-description.js
+++ b/lib/rules/require-meta-docs-description.js
@@ -45,6 +45,7 @@ module.exports = {
     return {
       Program() {
         const sourceCode = context.getSourceCode();
+        const { scopeManager } = sourceCode;
         const info = utils.getRuleInfo(sourceCode);
 
         if (info === null) {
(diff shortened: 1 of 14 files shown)


bmish:
Ready for review.

aladdin-add (approved):
LGTM, thanks! 💯

piranna:
great :-)

PR #253, opened 2022-06-20
Update `prefer-message-ids` to check for missing or unused messageIds

### LINKED ISSUE #57: Rule: valid usage of messageId
[ESLint v4.15.0](https://eslint.org/blog/2018/01/eslint-v4.15.0-released) introduced [`messageId`](https://eslint.org/docs/developer-guide/working-with-rules#messageids). Looking at this code sample:

```js
module.exports = {
  meta: {
    messages: {
      avoidName: "Avoid using variables named '{{ name }}'",
    },
  },
  create(context) {
    return {
      Identifier(node) {
        if (node.name === 'foo') {
          context.report({
            node,
            messageId: 'avoidName',
            data: {
              name: 'foo',
            },
          })
        }
      },
    }
  },
}
```

I would want a rule that:

- ensures that the `messageId` used in `context.report()` matches a key in `meta.messages` (in this example, `avoidName`)
- if `meta.messages[messageId]` contains a placeholder, it matches `data[placeholderName]` (in this example, `{{ name }}` matches `data.name`)

I'm wondering if the second bullet point should be added to [`eslint-plugin/no-unused-placeholders`](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-unused-placeholders.md) or as part of a new rule. I think the first case makes sense as its own rule.

What are your thoughts on this? I'd be happy to work on a pull request this week - looking for feedback on if this would be valuable and how to best include these features into this plugin. Thanks!

bmish:
Fixes #57.

TODO
- [ ] Fix / write more rule tests
- [ ] Write unit tests for helper functions
- [ ] Handle spreads / variables in `meta.messages`
- [ ] Share helper functions with #252
- [ ] Or should this be separate rules `no-missing-message-ids` and `no-unused-message-ids`?

Part of the V5 release (https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/230).

### DIFF
--- a/lib/rules/prefer-message-ids.js
+++ b/lib/rules/prefer-message-ids.js
@@ -21,9 +21,11 @@ module.exports = {
     fixable: null,
     schema: [],
     messages: {
+      foundMessage: 'Use `messageId` instead of `message`.',
       messagesMissing:
         '`meta.messages` must contain at least one violation message.',
-      foundMessage: 'Use `messageId` instead of `message`.',
+      missingMessage: '`meta.messages` is missing this messageId.',
+      unusedMessage: 'This messageId was never used.',
     },
   },
 
@@ -31,6 +33,7 @@ module.exports = {
     const sourceCode = context.getSourceCode();
     const info = utils.getRuleInfo(sourceCode);
 
+    const messageIdsUsed = new Set();
     let contextIdentifiers;
 
     // ----------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


bmish:
I've decided this deserves a separate rule `no-invalid-message-ids` (or could be `require-valid-messeage-ids`) which can be a recommended rule even for people who don't want always prefer message ids by enabling `prefer-message-ids`.

PR #254, opened 2022-06-21
Add new rules `no-missing-message-ids` and `no-unused-message-ids`

bmish:
Fixes https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/57.

Detects:

* Using a `messageId` that does not exist
* Leaving a message in `meta.messages` that is unused

These were implemented as separate rules because:
* Each is quite complicated on its own and it helps to be able to test them in isolation
* In the event of false positives, the user can just disable one of the rules but not the other

Note that I eliminated many possible false positive scenarios by testing these rules on the ESLint codebase.

TODO

- [x] Fix failing test case
- [x] Write unit tests for util functions
- [x] Add JSdoc for util functions
- [x] ~Rename to `require-valid-message-ids`~
- [x] Split into separate rules `no-missing-message-ids` and `no-unused-message-ids` since the unused message ID check can have false positives when variables or helper functions are used
- [x] Test on ESLint repository to detect edge cases
- [x] Ignore usage check when can't statically determine messageIds used like below examples:

```js
function reportEnd(node, messageId) {
    context.report({
        node,
        messageId
    });
}
```

- [x] Consider variable messageId:

```js
let messageId;
if (foo) {
  messageId = 'abc';
} else {
  messageId = 'def';
}

context.report({
  node,
  messageId
});
```

- [x] Handle conditional in suggestions

```js
context.report({
    node,
    messageId: "unexpectedRegExp",
    suggest: noFix ? [] : [{
        messageId: "replaceWithLiteral",
    }]
});
```

- [x] Try to handle some false positives in eslint-plugin-unicorn with dynamic object keys:

```js
const MESSAGE_ID = 'no-this-assignment';
const messages = {
	[MESSAGE_ID]: 'Do not assign `this` to `{{name}}`.',
};
```

After this is merged:

* Use util functions in #252

### DIFF
--- a/lib/rules/no-missing-message-ids.js
+++ b/lib/rules/no-missing-message-ids.js
@@ -1,0 +1,88 @@
+'use strict';
+
+const utils = require('../utils');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'disallow `messageId`s that are missing from `meta.messages`',
+      category: 'Rules',
+      recommended: false,
+      url: 'https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/tree/HEAD/docs/rules/no-missing-message-ids.md',
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      missingMessage: '`meta.messages` is missing this `messageId`.',
+    },
+  },
+
+  create(context) {
+    const sourceCode = context.getSourceCode();
+    const { scopeManager } = sourceCode;
+    const ruleInfo = utils.getRuleInfo(sourceCode);
+
+    const messagesNode = utils.getMessagesNode(ruleInfo, scopeManager);
+
+    let contextIdentifiers;
+
+    if (!messagesNode || messagesNode.type !== 'ObjectExpression') {
+      // If we can't find `meta.messages`, disable the rule.
+      return {};
+    }
+
+    return {
+      Program(ast) {
+        contextIdentifiers = utils.getContextIdentifiers(scopeManager, ast);
+      },
+
+      CallExpression(node) {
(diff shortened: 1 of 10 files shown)


lib/rules/no-missing-message-ids.js
```diff
@@ -0,0 +1,88 @@
...
+      description:
+        'disallow `messageId`s that are missing from `meta.messages`',
+      category: 'Rules',
+      recommended: false,
+      url: 'https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/tree/HEAD/docs/rules/no-missing-message-ids.md',
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      missingMessage: '`meta.messages` is missing this `messageId`.',
```
aladdin-add, line 23: could the message be clearer, e.g. `messageId "foo"`?

lib/rules/no-unused-message-ids.js
```diff
@@ -0,0 +1,113 @@
...
+    docs: {
+      description: 'disallow unused `messageId`s in `meta.messages`',
+      category: 'Rules',
+      recommended: false,
+      url: 'https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/tree/HEAD/docs/rules/no-unused-message-ids.md',
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      unusedMessage: 'This message is never used.',
```
aladdin-add, line 22: same here. :)

aladdin-add (approved):
Looks so great! 💯 

I've tested it in a few repos:
* eslint
* eslint-plugin-node
* eslint-plugin-react
* acorn (to find unexpected crashes on no-eslint-rule js)

It's working very well, thanks! 👍

lib/rules/no-unused-message-ids.js
bmish: Yes updated.

bmish:
@aladdin-add updated the messages and I also fixed some false positives affecting eslint-plugin-unicorn due to variable message object keys.

PR #258, opened 2022-07-07
breaking: add `recommended` rules `prefer-message-ids`, `prefer-output-null`, `no-missing-message-ids`, `no-unused-message-ids`

bmish:
Enable additional `recommended` rules:
   * [prefer-message-ids](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/prefer-message-ids.md): message IDs have become best practice, ESLint core enforces this rule already
   * [prefer-output-null](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/prefer-output-null.md): this will enforce upcoming ESLint v9 behavior from an approved [RFC](https://github.com/eslint/rfcs/tree/main/designs/2021-stricter-rule-test-validation#autofix-and-suggestion-output-must-differ-from-the-test-case-code)
   * https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/254

Remove from `recommended`:
* ~~[consistent-output](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/consistent-output.md): no longer necessary since ESLint v7 [validates](https://eslint.org/docs/user-guide/migrating-to-7.0.0#rule-tester-strict) that test cases provide `output` when they produce an autofix, and it's tedious to include `output: null` when not necessary, so we shouldn't require it~~

Part of v5 release (https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/230).


aladdin-add:
> [consistent-output](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/consistent-output.md): no longer necessary since ESLint v7 [validates](https://eslint.org/docs/user-guide/migrating-to-7.0.0#rule-tester-strict) that test cases provide output when they produce an autofix, and it's tedious to include output: null when not necessary, so we shouldn't require it.

it's no harm not to remove it - linting can find the error earilier?

ddzz (approved):

PR #277, opened 2022-08-03
Allow additional types of dynamic schemas in `require-meta-schema` rule

### LINKED ISSUE #273: `require-meta-schema` should allow variable schema
If someone wants to use a variable or function call to get the value for any of the meta properties, that should be allowed.

```js
{
  schema: baseRule.meta.schema
}
```

```js
{
  schema: getSchema()
}
```

Example: https://github.com/typescript-eslint/typescript-eslint/blob/ad412cdd74dc5619fbe4bf27c0a5eb5c5a4b65ca/packages/eslint-plugin/src/rules/brace-style.ts#L27

This issue is not specific to TypeScript.

bmish:
The goal is to ban obviously-invalid types like `Literal` (string/number) for `meta.schema`, while allowing any kind of variables/function calls/dynamic calculation of the schema.

Previously, a simple variable was allowed, but now these are also allowed:

```js
{
  schema: baseRule.meta.schema
}
```

```js
{
  schema: getSchema()
}
```

Fixes #273.

### DIFF
--- a/lib/rules/require-meta-schema.js
+++ b/lib/rules/require-meta-schema.js
@@ -104,7 +104,10 @@ module.exports = {
           hasEmptySchema = true;
         }
 
-        if (!['ArrayExpression', 'ObjectExpression'].includes(value.type)) {
+        if (
+          value.type === 'Literal' ||
+          (value.type === 'Identifier' && value.name === 'undefined')
+        ) {
           context.report({ node: value, messageId: 'wrongType' });
         }
       },
--- a/tests/lib/rules/require-meta-schema.js
+++ b/tests/lib/rules/require-meta-schema.js
@@ -63,6 +63,7 @@ ruleTester.run('require-meta-schema', rule, {
       `,
       parserOptions: { sourceType: 'module' },
     },
+    // Variable schema with array value.
     `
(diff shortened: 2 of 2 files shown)


lib/rules/require-meta-schema.js
```diff
@@ -104,7 +104,16 @@ module.exports = {
...
+        if (
+          ![
+            'ArrayExpression',
+            'CallExpression',
+            'ConditionalExpression',
+            'LogicalExpression',
+            'MemberExpression',
+            'ObjectExpression',
+          ].includes(value.type)
+        ) {
```
aladdin-add, line 116: just wondering we can explicitly say what we want to check - as more expressions could be added in future es.

e.g some new added expressions:
* ChainExpression
* LogicalAssignmentExpression

bmish: Yeah I'm worried about this too. I think I should update this to instead only check a few common invalid types like `Literal`.

bmish: Switched to only checking for literals / null / undefined to be safe.

aladdin-add (approved):

PR #286, opened 2022-08-11
Fix false positive with `no-unused-message-ids` from external violation reporting function

### LINKED ISSUE #283: False positive with `no-unused-message-ids` when violation reporting is done with external helper function
In this example, the `context.report()` calls are all in a separate helper file, so we can't detect that the `messageId`s are used.

We should ignore this situation where we can't find any `context.report()` calls.

https://github.com/typescript-eslint/typescript-eslint/blob/b6e5413fb12c2b1943a288b514f864c339380388/packages/eslint-plugin/src/rules/naming-convention.ts#L11

bmish:
Fixes #283.

When we haven't found any violation reporting code in a rule, we should assume the violation reporting is being handled in an external file/function, so don't flag any violations as they are likely to be false positives.

Changes:

* Adds variable `hasSeenViolationReport` for tracking whether we have seen any violation reporting code
* Renames existing variable to `hasSeenUnknownMessageId ` for clarity

### DIFF
--- a/lib/rules/no-unused-message-ids.js
+++ b/lib/rules/no-unused-message-ids.js
@@ -30,7 +30,8 @@ module.exports = {
 
     const messageIdsUsed = new Set();
     let contextIdentifiers;
-    let shouldPerformUnusedCheck = true;
+    let hasSeenUnknownMessageId = false;
+    let hasSeenViolationReport = false;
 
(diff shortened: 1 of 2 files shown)


aladdin-add (approved):
LGTM, thanks!

PR #299, opened 2022-09-10
chore: create `dependabot.yml`

MichaelDeBoey:
This will create automatic PRs to update both dependencies + actions used in workflows, so we won't need manual PRs like #298 anymore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: daily
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
   - package-ecosystem: npm
     directory: /
     schedule:
-      interval: daily
+      interval: weekly


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: daily
```
bmish, line 11: Can we do `weekly`? I'm a big fan of dependabot but I don't want to get spammed by updates every day.

```suggestion
      interval: weekly
```

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
```
bmish, line 6: Can we do `weekly`? I'm a big fan of dependabot but I don't want to get spammed by updates every day.

```suggestion
      interval: weekly
```

MichaelDeBoey: Since we're only using major version on actions, I think it would be better to keep `daily`, so we can take action asap

bmish: Fair enough.

bmish (approved):

PR #300, opened 2022-09-11
fix: update repo link

MichaelDeBoey:


package.json
```diff
@@ -28,12 +28,12 @@
   ],
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin.git"
+    "url": "https://github.com/eslint-community/eslint-plugin-eslint-plugin"
```
bmish, line 31: Should still use git URL: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#repository

```suggestion
    "url": "git@github.com:eslint-community/eslint-plugin-eslint-plugin.git"
```

MichaelDeBoey: In my experience it's not required to do so & the link you provided isn't saying it either. 🤔

bmish: The example from my link shows the URL ending with `.git`.

> repository
> Specify the place where your code lives. This is helpful for people who want to contribute. If the git repo is on GitHub, then the npm docs command will be able to find you.
> 
> Do it like this:
> 
> {
>   "repository": {
>     "type": "git",
>     "url": "https://github.com/npm/cli.git"
>   }
> }
> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It's for computers.

MichaelDeBoey: Oh yeah totally missed that. 🙈

It's not necessary though 🤷‍♂️

bmish (approved):

MichaelDeBoey:
@bmish It seems like this was not published automatically.
Is there something missing in order to release this?

Not really necessary asap, but would be good to have an automatic release system in place I think

bmish:
@MichaelDeBoey will create an issue to discuss this.

PR #312, opened 2022-10-09
Automate docs with eslint-doc-generator

bmish:
I built this CLI tool [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) for automating the generation of the README rules list and rule doc title/notices for ESLint plugins. It follows common documentation conventions from this and other top ESLint plugins and will help us standardize documentation across ESLint plugins (and generally improve the usability of our custom rules through better documentation and streamline the process of adding new rules).

This is a follow-up to my previous work to manually improve the documentation. eslint-doc-generator is significantly more robust compared to the previous documentation generator script/tests which I have now removed. It has additional functionality (for example, the rules list legend is auto-generated, and the rules list will show additional columns of information when applicable) and 100% test coverage.


MichaelDeBoey (approved):
I love this package! 😍

Would be a good one to add to the @eslint-community org if we would do this in all packages I think 🤔

Maybe we should wait for v1 until we do this though?

aladdin-add (approved):
👍

bmish:
Thanks @MichaelDeBoey! Yes I think it's a good goal to move eslint-doc-generator to @eslint-community after it reaches a v1 release. I am currently working to implement all the essential functionality and bug fixes in it so that I can introduce it to ~10 of the top ESLint plugins (and address feedback that I hear during this process).

bmish:
That's fair. But yeah, it shouldn't be a big deal, I'll be keeping this updated as I work toward v1.

PR #313, opened 2022-10-12
chore(CI): simplify testing strategy

MichaelDeBoey:


bmish:
Can you summarize what's changed?

MichaelDeBoey:
@bmish I brought this file in line with what I did for `@mysticatea/eslint-plugin` in https://github.com/eslint-community/mysticatea__eslint-plugin/pull/1

- Testing against all supported Node versions
- Testing against other platforms
- Testing against older supported ESLint versions
- Splitting the Lint job

.github/workflows/main.yml
```diff
@@ -1,48 +1,75 @@
...
-      with:
-        node-version: "18.x"
-    - run: npm install
-    - run: npm install --save-dev eslint@7
-    - run: npm test
+      - name: 📥 Install dependencies
+        run: npm install
+
+      - name: 📥 Install ESLint
+        run: npm install eslint@${{ matrix.eslint }}
```
bmish, line 68: Since this is a dev-dependency, shouldn't we technically use this to override the current version:
```suggestion
        run: npm install --save-dev eslint@${{ matrix.eslint }}
```

bmish:
This should fix one of the issues: #315

.github/workflows/main.yml
MichaelDeBoey: I would use `--no-save` instead, WDYT?

bmish: Haven't used that but seems fine.

MichaelDeBoey: In fact, it doesn't matter actually, as we're not saving the updated version anywhere
It's only used to use the specified ESLint version in our tests

I would just leave it as-is tbh.

bmish: I disagree with that because we should be consistent about how eslint is installed. Simply using `npm install` will result in it being specified in two places during CI (both devDependencies and dependencies). This would be very confusing during debugging and result in us having potentially different versions.

I think this might also be why CI is broken right now:

```
Error: Cannot find module 'eslint'
Require stack:
- /home/runner/work/eslint-plugin-eslint-plugin/eslint-plugin-eslint-plugin/tests/lib/rules/consistent-output.js
- ```

MichaelDeBoey: Changed 👍

PR #314, opened 2022-10-12
chore(CI): add automatic release

### LINKED ISSUE #301: Consider automatic releases
_Originally posted by @MichaelDeBoey in https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/300#issuecomment-1243013325_

We currently use release-it to perform releases via a CLI command. I'm personally open to setting up the package to automatically release based on semantic commits. Any thoughts or recommendations on common/standard ways to do this?

MichaelDeBoey:
@eslint-community/eslint-tsc In order to make the automatic release work, we should add the `NPM_TOKEN` secret in [the Action secret settings](https://github.com/organizations/eslint-community/settings/secrets/actions).

I think it's best to set a PAT for [the `eslint-community-bot` npm user](https://www.npmjs.com/~eslint-community-bot) btw.

---

This branch depends on #313

Closes #301

---

This PR is inspired by what I did for `@mysticatea/eslint-plugin` in https://github.com/eslint-community/mysticatea__eslint-plugin/pull/8

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,14 @@
 name: CI
 on:
   push:
+    branches:
+      # default semantic-release branches
+      - +([0-9])?(.{+([0-9]),x}).x
+      - main
+      - next
+      - next-major
+      - beta
+      - alpha
   pull_request:
 
 jobs:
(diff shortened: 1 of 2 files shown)


nzakas:
I added an npm automation token as NPM_TOKEN at the organization level. To be safe it’s not enabled for any repos. You can enable it just for the repos that need it.

bmish:
@MichaelDeBoey how will the automatic releases work? When will they happen? And what if we want to do a single release with multiple changes?

MichaelDeBoey:
> I added an npm automation token as NPM_TOKEN at the organization level.

@nzakas Thanks! 🙏

---

> how will the automatic releases work? When will they happen?

@bmish [`semantic-release`](https://github.com/semantic-release/semantic-release) will take care of the release on every commit that's done to the `main`/`next`/`alpha`/`beta` branch
It will look at the commit message  & release a patch version for `fix`, a minor version for `feat` & a major version for every commit that has a `BREAKING CHANGE` in their body
More info on how they work, can be found at https://github.com/semantic-release/semantic-release#how-does-it-work

> what if we want to do a single release with multiple changes?

You can create pre-releases when pushing to the `alpha`/`beta` branch. Each commit will still be released, but not for people who install the package normally, only for people who use the `-alpha.X`/`-beta.X` suffix or `alpha`/`beta` token (like `npm install eslint-plugin-eslint-plugin@alpha`).
Have a look at how this is working for `testing-library-eslint-config`:
- https://github.com/testing-library/eslint-plugin-testing-library/releases
- https://www.npmjs.com/package/eslint-plugin-testing-library?activeTab=versions

bmish:
Is this compatible with the conventional commits we currently use? https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/221

MichaelDeBoey:
> Should we add the github badge so people know things will be released automatically?

Was going to cleanup the badges in a separate PR

> Should we get rid of `release-it` at this point to avoid having multiple systems in place for releasing?

Indeed a good idea to remove it

> Is this compatible with the conventional commits we currently use?

Yes it very much is!
The docs I referenced are even mentioning [`commitizen`](https://github.com/commitizen/cz-cli) & [`commitlint`](https://github.com/conventional-changelog/commitlint)

> > Tools such as [commitizen](https://github.com/commitizen/cz-cli) or [commitlint](https://github.com/conventional-changelog/commitlint) can be used to help contributors and enforce valid commit messages.

Issue #269, opened 2022-07-12
V6 Planned Changes


bmish:
Starting a list of my desired breaking changes, feedback welcome. We can perform this release once we have compiled enough changes. We should watch out for any breaking changes needed for [ESLint v9](https://github.com/eslint/eslint/projects/9) in the coming months. 

Planned:
* Enable additional `recommended` rules:
   * TODO
* Change rule options
   * #320
   * #430
   * #402
   * Enable `catchNoFixerButFixableProperty` by default on `require-meta-fixable` rule
      * After making fixes to reduce false positives like we did with `require-meta-has-suggestions`
* ESLint 9
   * Change option default for `requireSchemaPropertyWhenOptionless` to `false` on `require-meta-schema`
* Drop support for ESLint v7 and old versions of Node
* Future
   * Stop supporting function-style rules which will no longer be supported in ESLint v9. Actively removing support would be a lot of work so I'm in no hurry to do this an we can push it off indefinitely.
* TODO

Maybe:
* Convert package to [ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm) or TypeScript?

For reference, V5 was release 2022-07 ([CHANGELOG](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/releases/tag/v5.0.0), discussion in #230).

aladdin-add:
one thing to discuss:

the flat confg supports was added in [v5.1.0](https://github.com/eslint-community/eslint-plugin-eslint-plugin/releases/tag/v5.1.0), but it's required to access internal files: `require("eslint-plugin-eslint-plugin/configs/recommended")`.  we can just convert the configs to flat if eslintrc supports can be dropped in v6.

PR #320, opened 2022-10-21
feat!: Add missing `deprecated`/`replacedBy` properties to `meta-property-ordering` rule

bmish:
The `deprecated`, `replacedBy` rule metadata properties were mentioned in the rule doc but not actually enforced by the rule.

I have inserted these properties into the default list right after `fixable`/`hasSuggestions`. I'm leaving `schema`/`messages` last in the list since these are potentially large objects/arrays, and we should keep the essential, concise/one-line properties together.

This is a breaking change that can go in with #269.


docs/rules/meta-property-ordering.md
```diff
@@ -12,7 +12,7 @@ This rule enforces that meta properties of a rule are placed in a consistent ord
 
 This rule has an array option:
 
-* `['type', 'docs', 'fixable', 'hasSuggestions', 'schema', 'messages', 'deprecated', 'replacedBy']` (default): The order that the properties of `meta` should be placed in.
+* `['type', 'docs', 'fixable', 'hasSuggestions', 'deprecated', 'replacedBy', 'schema', 'messages']` (default): The order that the properties of `meta` should be placed in.
```
aladdin-add, line 15: imo, the two props should be highlighted, what do you think change them to be in the 1st place?

bmish: Sure, I would be open to that.

aladdin-add (approved):

PR #335, opened 2023-01-03
chore: add `eslint-remote-tester`

bmish:
This runs all our rules on top ESLint plugin repos to ensure that they don't crash. It's like smoke testing. 

https://github.com/AriPerkkio/eslint-remote-tester

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ node_modules/
 npm-debug.log
 yarn.lock
 .eslintcache
+
+# eslint-remote-tester
+.cache-eslint-remote-tester
+eslint-remote-tester-results
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,4 @@ yarn.lock
 .eslintcache
 
 # eslint-remote-tester
-.cache-eslint-remote-tester
 eslint-remote-tester-results
--- a/eslint-remote-tester.config.js
+++ b/eslint-remote-tester.config.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const { getPathIgnorePattern } = require('eslint-remote-tester-repositories');
-
 /** @type {import('eslint-remote-tester').Config} */
 module.exports = {
   /** Repositories to scan */
@@ -37,11 +35,8 @@ module.exports = {
     'typescript-eslint/typescript-eslint',
   ],
 
-  /** Optional pattern used to exclude paths */
-  pathIgnorePattern: getPathIgnorePattern(),
-
   /** Extensions of files under scanning */
-  extensions: ['js', 'ts'],
+  extensions: ['js', 'mjs', 'cjs', 'ts', 'mts', 'cts'],
 
   /** Optional boolean flag used to enable caching of cloned repositories. For CIs it's ideal to disable caching. Defaults to true. */
   cache: false,
@@ -49,5 +44,12 @@ module.exports = {
   /** ESLint configuration */
   eslintrc: {
     extends: ['plugin:eslint-plugin/all'],
+
+    overrides: [
+      {
+        files: ['*.ts', '*.mts', '*.cts'],
+        parser: '@typescript-eslint/parser',
+      },
+    ],
   },
 };
--- a/eslint-remote-tester.config.js
+++ b/eslint-remote-tester.config.js
@@ -1,0 +1,52 @@
+'use strict';
+
+const { getPathIgnorePattern } = require('eslint-remote-tester-repositories');
+
+module.exports = {
+  /** Repositories to scan */
+  repositories: [
+    // A few dozen top ESLint plugins.
+    'Intellicode/eslint-plugin-react-native',
+    'JoshuaKGoldberg/eslint-plugin-expect-type',
+    'SonarSource/eslint-plugin-sonarjs',
+    'avajs/eslint-plugin-ava',
+    'cypress-io/eslint-plugin-cypress',
+    'dangreenisrael/eslint-plugin-jest-formatting',
+    'ember-cli/eslint-plugin-ember',
+    'emberjs/eslint-plugin-ember-internal',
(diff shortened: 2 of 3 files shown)


aladdin-add (approved):
LGTM~

MichaelDeBoey:
CC @AriPerkkio Can you review this please?

eslint-remote-tester.config.js
```diff
@@ -0,0 +1,52 @@
...
+  pathIgnorePattern: getPathIgnorePattern(),
+
+  /** Extensions of files under scanning */
+  extensions: ['js', 'ts'],
+
+  /** Optional boolean flag used to enable caching of cloned repositories. For CIs it's ideal to disable caching. Defaults to true. */
+  cache: false,
+
+  /** ESLint configuration */
+  eslintrc: {
```
AriPerkkio, line 49: Should the typescript parser also be included here?

https://github.com/jsx-eslint/eslint-plugin-react/blob/a5f70658d7d68a9d76c3a23b1fa2e1cfe02438ee/test/eslint-remote-tester.config.js#L24-L29

package.json
```diff
@@ -63,6 +64,8 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^4.2.1",
     "eslint-plugin-unicorn": "^44.0.0",
+    "eslint-remote-tester": "^3.0.0",
+    "eslint-remote-tester-repositories": "^1.0.0",
```
AriPerkkio, line 68:
```suggestion
```

AriPerkkio (changes requested):

eslint-remote-tester.config.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+const { getPathIgnorePattern } = require('eslint-remote-tester-repositories');
```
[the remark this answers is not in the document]
MichaelDeBoey: Should we use the `repositories` from `eslint-remote-tester-repositories` instead?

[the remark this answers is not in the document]
AriPerkkio: Currently it seems that this setup runs tests against a small set of ESLint plugin repositories, as part of CI. This approach is good, and I think `eslint-plugin-unicorn` has similar setup. They also use a second smoke test workflow for a ~5 hour test run against randomly picked ~10k repositories. 

So which ever approach is ok. Do you want to run these tests quickly for every CI run, or schedule them to run once a week for large set of repositories. Or both?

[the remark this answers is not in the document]
bmish: Thanks for explaining both approaches.

I figured we could start by running against a set of ESLint plugins, since those repos are of course more relevant and likely to trigger our rules than any other random repository. The current set is small so it's quick to run during CI.

If there was a desire to run on a very large set of repos but less frequently, I'm fine with that too.

eslint-remote-tester.config.js
bmish: Good point, added.

eslint-remote-tester.config.js
```diff
@@ -0,0 +1,52 @@
...
+    'sindresorhus/eslint-plugin-unicorn',
+    'square/eslint-plugin-square',
+    'storybookjs/eslint-plugin-storybook',
+    'testing-library/eslint-plugin-jest-dom',
+    'testing-library/eslint-plugin-testing-library',
+    'typescript-eslint/typescript-eslint',
+  ],
+
+  /** Optional pattern used to exclude paths */
+  pathIgnorePattern: getPathIgnorePattern(),
```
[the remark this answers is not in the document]
bmish: Thanks!

eslint-remote-tester.config.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+const { getPathIgnorePattern } = require('eslint-remote-tester-repositories');
+
+module.exports = {
```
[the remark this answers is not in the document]
bmish: Nice!

eslint-remote-tester.config.js
```diff
@@ -0,0 +1,55 @@
...
+    'testing-library/eslint-plugin-jest-dom',
+    'testing-library/eslint-plugin-testing-library',
+    'typescript-eslint/typescript-eslint',
+  ],
+
+  /** Extensions of files under scanning */
+  extensions: ['js', 'mjs', 'cjs', 'ts', 'mts', 'cts'],
+
+  /** Optional boolean flag used to enable caching of cloned repositories. For CIs it's ideal to disable caching. Defaults to true. */
+  cache: false,
```
bmish, line 42: @AriPerkkio I was a bit confused by this `cache` option. I some some other repos disabling it. Can I just let it use the default? Or why is the default value not sufficient?

AriPerkkio: This flag controls whether a cloned repository should be removed from file system after it was linted. The default value `true` is ideal for local development, where developer may have 100s of repositories cached in file system, and they keep linting against them after modifying rule implementation frequently. Cloning repositories is slow, so caching them makes testing fast. Default values of this tool are good for local development, so some changes may be required when targeting CIs. 

In this case, as there are not that many repositories listed, it might be OK to leave this as `true`. Github CIs have 15-20GB file systems on free tier. When linting ~10K repositories, the caching must be disabled so that we don't run out of disk space after couple of hours.

.gitignore
```diff
@@ -5,3 +5,7 @@ node_modules/
 npm-debug.log
 yarn.lock
 .eslintcache
+
+# eslint-remote-tester
+.cache-eslint-remote-tester
```
AriPerkkio, line 10: Cache is nowadays stored under `node_modules` at `node_modules/eslint-remote-tester/.cache-eslint-remote-tester/`, so no need to specify it in `.gitignore`.

```suggestion
```

AriPerkkio (approved):

PR #337, opened 2023-01-20
perf: use improved comparison algorithm in `no-identical-tests` rule

marvinhagemeister:
Spent some time this evening profiling eslint on the eslint repo. Noticed that this rule seems to take a little longer than I would've guessed and spend a bit of time with @jviide improving the equality algorithm used for this rule. The result is that the changes in this PR make the rule ~10x faster.

| Before | After |
|--:|--:|
| 453ms | **46ms** |

### DIFF
--- a/lib/rules/no-identical-tests.js
+++ b/lib/rules/no-identical-tests.js
@@ -38,48 +38,28 @@ module.exports = {
     // Helpers
     // ----------------------------------------------------------------------
     /**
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
Alright it's ready now, fixed a typo in the commit message 🙈

bmish (approved):
Awesome, thanks!

PR #350, opened 2023-04-26
chore: add `format` workflow

MichaelDeBoey:


bmish:
Can you explain the motivation? Why do we want CI to attempt to create a formatting commit? Every PR is already required to pass formatting/linting before being merged.

MichaelDeBoey:
@bmish formatting is currently not enforced, hence the new workflow

bmish:
It's enforced via linting since we use the `prettier/recommended` config. And regardless, I think we should enforce it during CI in PRs instead of having CI create its own commits on main to fix formatting,

MichaelDeBoey:
@bmish The reason why I set it on `main` is because it can be auto-fixed, so we don't need to keep PRs open just for these kind of things

aladdin-add:
I agreed with @bmish, and we don't have many PRs for formatting.

PR #355, opened 2023-06-13
Tweak violation message in `prefer-message-ids` rule

### LINKED ISSUE #354: Error: Use `messageId` instead of `message` is incorrect for field `desc`
This error will show if the suggestion has `desc` instead of `messageId`

The error message is misleading

I am willing to submit a PR

nopeless:
Fixes #354

### DIFF
--- a/lib/rules/prefer-message-ids.js
+++ b/lib/rules/prefer-message-ids.js
@@ -13,7 +13,7 @@ module.exports = {
     type: 'problem',
     docs: {
       description:
-        'require using `messageId` instead of `message` to report rule violations',
+        'require using `messageId` instead of `message` or `desc` to report rule violations',
       category: 'Rules',
       recommended: true,
       url: 'https://github.com/eslint-community/eslint-plugin-eslint-plugin/tree/HEAD/docs/rules/prefer-message-ids.md',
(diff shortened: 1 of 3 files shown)


nopeless:
@bmish

lib/rules/prefer-message-ids.js
```diff
@@ -23,7 +23,10 @@ module.exports = {
     messages: {
       messagesMissing:
         '`meta.messages` must contain at least one violation message.',
-      foundMessage: 'Use `messageId` instead of `message`.',
+      // Currently the codebase does not distinguish between `message` and `desc`.
+      // `utils.collectReportViolationAndSuggestionData` returns `message` for both.
+      foundMessage:
+        'Use `messageId` instead of `message` (for reports) or `desc` (for suggestions).',
```
bmish, line 29:
```suggestion
        'Use `messageId` instead of `message` (for violations) or `desc` (for suggestions).',
```

lib/rules/prefer-message-ids.js
```diff
@@ -23,7 +23,10 @@ module.exports = {
     messages: {
       messagesMissing:
         '`meta.messages` must contain at least one violation message.',
-      foundMessage: 'Use `messageId` instead of `message`.',
+      // Currently the codebase does not distinguish between `message` and `desc`.
+      // `utils.collectReportViolationAndSuggestionData` returns `message` for both.
```
bmish, line 27:
```suggestion
```

I think we can omit this comment. It's helpful context for the PR description but we probably don't need a comment for it.

bmish (commented):
Please include `Fixes #354` (without backticks) in the PR description to link the PR and issue.

nopeless:
rebased, commmit history clean @bmish

bmish (approved):
Thanks!

PR #387, opened 2023-10-12
docs: auto-generate some rule option lists with eslint-doc-generator

bmish:
Moves information about rule options into rule schemas so rule doc option lists can be autogenerated.

Note that I have only added these auto-generated option lists to rules with simple object-based options. Other types of options (enums/arrays/etc) aren't supported yet.

This feature shipped in:
* https://github.com/bmish/eslint-doc-generator/releases/tag/v1.5.0
* https://github.com/bmish/eslint-doc-generator/pull/480

### DIFF
--- a/docs/rules/require-meta-docs-description.md
+++ b/docs/rules/require-meta-docs-description.md
@@ -49,9 +49,13 @@ module.exports = {
 
 ## Options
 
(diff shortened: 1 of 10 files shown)


aladdin-add (approved):
LGTM, thanks!

PR #406, opened 2023-12-12
feat!: change test-case-property-ordering to add supports languageOptions

### LINKED ISSUE #402: Support flat config properties for `test-case-property-ordering`
`languageOptions` has replaced `parserOptions` in flat config. But we can keep them both side-by-side in our ordered list of test case properties so that we can support both kinds of config.

Related:
* https://github.com/eslint/eslint/pull/17819#discussion_r1422591180

aladdin-add:
fixes #402

### DIFF
--- a/lib/rules/test-case-property-ordering.js
+++ b/lib/rules/test-case-property-ordering.js
@@ -45,10 +45,10 @@ module.exports = {
       'output',
       'options',
       'parser',
-      'languageOptions', // used in eslint flat config
-      'parserOptions', // used in eslintrc config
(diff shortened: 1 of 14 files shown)


aladdin-add:
I don't think it will be a breaking change - `parserOptions` & `langugeOptions` cannot co-exists.

bmish:
I believe it's a breaking change as anyone with flat rule tests that use `languageOptions` will suddenly have a new ordering of this property enforced.

bmish (approved):

PR #433, opened 2024-02-05
feat: add no-property-in-node rule

### LINKED ISSUE #326: Rule request: don't use the in operator on nodes
Continuing conversation from here https://github.com/typescript-eslint/typescript-eslint/pull/5610#discussion_r1014918727: We (typescript-eslint) generally try not to use string `in` checking pattern on AST nodes, as it's a little imprecise. We normally check for specific types.

```diff
- if ('async' in node && node.async) {
+ if (isFunctionLike(node) && node.async) {
```

How about making this an ESLint rule?

cc @eliasm307 since we were chatting 🙂

JoshuaKGoldberg:
Fixes #326.


README.md
```diff
@@ -3,6 +3,7 @@
 An ESLint plugin for linting ESLint plugins. Rules written in CJS, ESM, and TypeScript are all supported.
 
 <!-- vscode-markdown-toc -->
+
```
JoshuaKGoldberg, line 6: [Question] I kept having to undo Prettier changes to the README.md 😅 . Would you mind a separate issue around having it eslint-doc-generator and Prettier not come into conflict?

README.md
```diff
@@ -69,63 +70,84 @@ module.exports = [
...
-|   | `rules-recommended` | enables all recommended rules that are aimed at linting ESLint rule files |
-|   | `tests-recommended` | enables all recommended rules that are aimed at linting ESLint test files |
-|   | `all`               | enables all rules in this plugin                                          |
-|   | `rules`             | enables all rules that are aimed at linting ESLint rule files             |
-|   | `tests`             | enables all rules that are aimed at linting ESLint test files             |
+|         | Name                                             | Description                                                                                                                                                            |
+| :-- | :------------------------- | :--------------------------------------------------------------------------------------- |
+| ✅    | `recommended`                            | enables all recommended rules in this plugin, excluding those requiring type information |
+|         | `recommended-type-checked` | enables all recommended rules in this plugin, including those requiring type information |
```
JoshuaKGoldberg, line 145: [Question] I don't _love_ that this doesn't have its own emoji 😞 is there some standard practice that can be done?

README.md
bmish: Sure! https://github.com/bmish/eslint-doc-generator#prettier

README.md
bmish: Yes I would just give it a different color checkmark like [☑️](https://emojipedia.org/check-box-with-check) in `.eslint-doc-generatorrc.js`.

lib/rules/no-property-in-node.js
```diff
@@ -0,0 +1,58 @@
...
+
+const typedNodeSourceFileTesters = [
+  /@types[/\\]estree[/\\]index\.d\.ts/,
+  /@typescript-eslint[/\\]types[/\\]dist[/\\]generated[/\\]ast-spec\.d\.ts/,
+];
+
+/**
+ * @param {import('typescript').Type} type
+ * @returns Whether the type seems to include a known ESTree or TSESTree AST node.
+ */
```
bmish, line 14: I'm a bit unfamiliar with some of the utils and concepts here so maybe including an example in the comment for this function would help explain what it's doing?

docs/rules/no-property-in-node.md
```diff
@@ -0,0 +1,38 @@
...
+Examples of **correct** code for this rule:
+
+```ts
+/* eslint eslint-plugin/no-property-in-node: error */
+
+declare const node: TSESTree.Parameter;
+
+if (node.type !== TSESTree.AST_NODE_TYPES.TSParameterProperty) {
+  node.optional;
+}
```
bmish, line 37: I feel like this could benefit from a more realistic / real-world / relatable example of correctly checking the node type before using its properties.

docs/rules/no-property-in-node.md
```diff
@@ -0,0 +1,38 @@
...
+💼 This rule is enabled in the `recommended-type-checked` [config](https://github.com/eslint-community/eslint-plugin-eslint-plugin#presets).
+
+💭 This rule requires type information.
+
+<!-- end auto-generated rule header -->
+
+When working with a node of type `ESTree.Node` or `TSESTree.Node`, it can be tempting to use the `'in'` operator to narrow the node's type.
+`'in'` narrowing is susceptible to confusing behavior from quirks of ASTs, such as node properties sometimes being omitted from nodes and other times explicitly being set to `null` or `undefined`.
+
+Using direct property checks is generally considered preferable.
```
bmish, line 12: "direct property checks" seems ambiguous and not really clear enough to be actionable. Shouldn't this suggest something about explicitly checking the node's type rather than simply checking for the presence of a property?

README.md
bmish, line 6: Looks like the table of contents needs to be updated which will be automatic if you have this extension:

Name: Markdown All in One
Id: yzhang.markdown-all-in-one
Description: All you need to write Markdown (keyboard shortcuts, table of contents, auto preview and more)
Version: 3.6.2
Publisher: Yu Zhang
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one

tests/lib/rules/no-property-in-node.js
```diff
@@ -0,0 +1,111 @@
...
+          column: 13,
+          line: 5,
+          endColumn: 24,
+          endLine: 5,
+          messageId: 'in',
+        },
+      ],
+    },
+  ],
+  valid: [
```
bmish, line 88: Can we have some valid examples that do correct type checking `node.type === ...` or just basic property access `node.a`.

lib/index.js
```diff
@@ -16,7 +16,9 @@ const PLUGIN_NAME = packageMetadata.name.replace(/^eslint-plugin-/, '');
 
 const configFilters = {
   all: () => true,
-  recommended: (rule) => rule.meta.docs.recommended,
+  recommended: (rule) =>
+    rule.meta.docs.recommended && !rule.meta.docs.requiresTypeChecking,
+  'recommended-type-checked': (rule) => rule.meta.docs.recommended,
```
bmish, line 21: FYI @aladdin-add on this new config.

tests/lib/rules/no-property-in-node.js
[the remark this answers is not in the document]
JoshuaKGoldberg: Heh, yeah, I think I'm the only person who does this. It's because I use `eslint-plugin-perfectionist` to sort everything alphabetically in my repos. I'll switch 🙂

README.md
JoshuaKGoldberg: Err, which command should I run?

bmish: Normally it just updates it for you when you save the file.

JoshuaKGoldberg: #434 -> #435 🚀

lib/rules/no-property-in-node.js
JoshuaKGoldberg: Does 3575644 answer what you're looking for?

bmish: Thanks!

README.md
JoshuaKGoldberg: Well, ah, this is what it updated me to.

README.md
```diff
@@ -4,12 +4,18 @@ An ESLint plugin for linting ESLint plugins. Rules written in CJS, ESM, and Type
 
 <!-- vscode-markdown-toc -->
 
-- [Installation](#Installation)
-- [Usage](#Usage)
-- [Rules](#Rules)
-- [Presets](#Presets)
-  - [Semantic versioning policy](#Semanticversioningpolicy)
-  - [Preset usage](#Presetusage)
+- [eslint-plugin-eslint-plugin   ](#eslint-plugin-eslint-plugin---)
```
bmish, line 7: Hmm this top-level title was supposed to be ignored.

package.json
```diff
@@ -40,8 +40,10 @@
   },
   "homepage": "https://github.com/eslint-community/eslint-plugin-eslint-plugin#readme",
   "dependencies": {
+    "@typescript-eslint/utils": "^6.20.0",
```
JoshuaKGoldberg, line 43: https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/7789482239/job/21241206507?pr=433:

```plaintext
> nyc --all --check-coverage --include lib mocha tests --recursive


/home/runner/work/eslint-plugin-eslint-plugin/eslint-plugin-eslint-plugin/node_modules/@typescript-eslint/scope-manager/dist/referencer/ClassVisitor.js:123
        withMethodDecorators ||=
                             ^^^

SyntaxError: Unexpected token '||='
    at wrapSafe (internal/modules/cjs/loader.js:102[9](https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/7789482239/job/21241206507?pr=433#step:5:10):16)
```

Is this blocked on dropping Node 14 support then?

package.json
```diff
@@ -55,7 +57,10 @@
     "@eslint/eslintrc": "^2.0.2",
     "@eslint/js": "^8.37.0",
     "@release-it/conventional-changelog": "^4.3.0",
-    "@typescript-eslint/parser": "^5.36.2",
+    "@types/eslint": "^8.56.2",
+    "@types/estree": "^1.0.5",
+    "@typescript-eslint/parser": "^6.20.0",
+    "@typescript-eslint/rule-tester": "^6.20.0",
```
JoshuaKGoldberg, line 63: https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/7789482239/job/21241205874?pr=433:

```plaintext
> nyc --all --check-coverage --include lib mocha tests --recursive


Error: Cannot find module 'eslint/use-at-your-own-risk'
Require stack:
- /home/runner/work/eslint-plugin-eslint-plugin/eslint-plugin-eslint-plugin/node_modules/@typescript-eslint/rule-tester/dist/utils/config-validator.js
```

Is this blocked on dropping ESLint 7 support then?

package.json
```diff
@@ -55,7 +57,10 @@
     "@eslint/eslintrc": "^2.0.2",
     "@eslint/js": "^8.37.0",
     "@release-it/conventional-changelog": "^4.3.0",
-    "@typescript-eslint/parser": "^5.36.2",
+    "@types/eslint": "^8.56.2",
+    "@types/estree": "^1.0.5",
+    "@typescript-eslint/parser": "^6.20.0",
```
JoshuaKGoldberg, line 62: Is there a strategy I should take around these `devDependencies` not being recognized by `node/no-missing-require`? 

```plaintext
$ npm run lint:js

> eslint-plugin-eslint-plugin@5.2.1 lint:js
> eslint --cache --ignore-pattern "**/*.md" .


/Users/josh/repos/eslint-plugin-eslint-plugin/lib/rules/no-property-in-node.js
  3:33  error  "@typescript-eslint/utils" is not found  node/no-missing-require

/Users/josh/repos/eslint-plugin-eslint-plugin/tests/lib/rules/no-property-in-node.js
  3:32  error  "@typescript-eslint/rule-tester" is not found  node/no-missing-require

/Users/josh/repos/eslint-plugin-eslint-plugin/tests/lib/rules/require-meta-docs-description.js
  298:27  error  "@typescript-eslint/parser" is not found  node/no-missing-require

/Users/josh/repos/eslint-plugin-eslint-plugin/tests/lib/utils.js
  10:40  error  "@typescript-eslint/parser" is not found  node/no-missing-require

✖ 4 problems (4 errors, 0 warnings)
```

Left to my own devices I'd try adding an override for test files that sets `settings.allowedModules`. And maybe file an issue on `eslint-plugin-n` to add a rule option for pulling in `devDependency`/`peerDependencies`/etc.

README.md
bmish: Can you add `<!-- omit from toc -->` to the end of the title line?

https://github.com/yzhang-gh/vscode-markdown#table-of-contents

package.json
bmish: Not immediately sure but a separate PR for `eslint-plugin-n` is welcome.

README.md
JoshuaKGoldberg: That did it, thanks!

package.json
bmish: Is there an older version of the dependency we can use that will still work for Node 14?

In terms of dropping for old Node/ESLint, I'd be open to doing this as part of a major version soon, although there's some work for that in #269.

bmish (approved):
Awesome, thank you!

lib/index.js
aladdin-add: SGTM! 👍

bmish:
@aladdin-add let's release this as a minor version. I'll leave it for you to review or merge and then conduct the minor release.

lib/index.js
JoshuaKGoldberg, line 1: From https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/7794182409/job/21255096639?pr=433:

```plaintext
Results:
Repository: eslint/eslint
Rule: eslint-plugin/no-property-in-node
Message: You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
Occurred while linting /home/runner/work/eslint-plugin-eslint-plugin/eslint-plugin-eslint-plugin/node_modules/eslint-remote-tester/.cache-eslint-remote-tester/eslint/eslint/docs/src/assets/js/focus-visible.js:41
Rule: "eslint-plugin/no-property-in-node"
Path: eslint/eslint/docs/src/assets/js/focus-visible.js
Link: https://github.com/eslint/eslint/blob/HEAD/docs/src/assets/js/focus-visible.js#L41
```

What would you like to do here? The first ideas that come to mind for me are:

* Disable type-checked rules in `all` & add an `all-type-checked` config?
* Disable `no-property-in-node` for specifically `eslint/eslint`, since it's the only one without a TSConfig?

aladdin-add: > Disable type-checked rules in all & add an all-type-checked config?

I slightly prefer this. just like`recommended` and `recommended-type-checked`. @bmish wdyt?

aladdin-add: the rule is also `rules-recommended`? So also need `rules-recommended-type-checked`...., so many presets  doesn't seem like a good idea.😅

JoshuaKGoldberg: Yeah 😓 ... another possibility for flat configs specifically could be to make the configs be a function with properties set on it. So you could do either of:

```js
export default [
  eslintPlugin.configs.recommended,
  eslintPlugin.configs.recommended({ typeChecked: true }),
]
```

The only prior art I know of this strategy is https://github.com/azat-io/eslint-plugin-perfectionist/issues/90.

aladdin-add: maybe just adding `recommended: false` in this PR. let's consider adding the configs in another PR. thoughts?

aladdin-add: `recommended` config follows the eslint's [semantic-versioning-policy](https://github.com/eslint/eslint?tab=readme-ov-file#semantic-versioning-policy), while `all`, `rule-all`, `test-all` does not.

package.json
JoshuaKGoldberg: Ah, removing the dependencies resolved the issue. Yay!

aladdin-add (approved):
Looks great, thanks! 💯

tests/lib/rules/no-property-in-node.js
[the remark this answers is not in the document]
StyleShit: Well, you're not the only one 😄 
https://github.com/microsoft/rushstack/blob/fe1cbf4e3f325e4d55c4c02316715999b72b4d20/eslint/eslint-plugin/src/test/typedef-var.test.ts

(an example I saw yesterday and had the same thought as bmish)

README.md
```diff
@@ -75,56 +79,59 @@ module.exports = [
...
-|     | `tests-recommended` | enables all recommended rules that are aimed at linting ESLint test files |
-|     | `all`               | enables all rules in this plugin                                          |
-|     | `rules`             | enables all rules that are aimed at linting ESLint rule files             |
-|     | `tests`             | enables all rules that are aimed at linting ESLint test files             |
+|     | Name                | Description                                                                  |
+| :-- | :------------------ | :--------------------------------------------------------------------------- |
+| ✅  | `recommended`       | enables all recommended rules in this plugin                                 |
+|     | `rules-recommended` | enables all recommended rules that are aimed at linting ESLint rule files    |
+|     | `tests-recommended` | enables all recommended rules that are aimed at linting ESLint test files    |
+|     | `all`               | enables all rules in this plugin, including those requiring type information |
```
bmish, line 131: looks like this should be `excluding`

lib/index.js
bmish: Duplicating every config when we have so many is not great. What about just providing a `type-checked` config with just this single rule that you can mix and match with the others?

README.md
JoshuaKGoldberg: dangit, sorry, copy & paste

lib/index.js
JoshuaKGoldberg: Heh, that's what we used to do in typescript-eslint. But then we found that most folks only enabled one of the configs. And it was a bit irksome to have to enable two.

https://github.com/typescript-eslint/typescript-eslint/issues/5204 -> https://typescript-eslint.io/blog/announcing-typescript-eslint-v6#reworked-configuration-names is what we ended up with.

bmish: Continuing discussion in: 
* https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/445

PR #447, opened 2024-02-17
feat: add require-meta-docs-recommended rule

### LINKED ISSUE #446: Rule: require a meta.docs.recommended
Similar to `require-meta-docs-description` and `require-meta-docs-url`: it's common for plugin rules to specify to specify a `meta.docs.description` value. For example, in `eslint-plugin-eslint-plugin`: https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/b8ed4caa9b0ffeb0e12c44824ccee5428345e90b/lib/rules/consistent-output.js#L22

Not every plugin enforces a `meta.docs.recommended` & correspondingly named ruleset, so I'm not sure whether this should be enabled by default in this plugin's `recommended` ruleset. But it'd be nice to at least have the rule available, I think.

Filing this as a continuation of https://github.com/eslint/eslint/pull/18005#discussion_r1458066983.  At least ESLint core does this ad hoc.

JoshuaKGoldberg:
Fixes #446.

Doesn't include any options, as the internal core ESLint `no-invalid-meta` rule doesn't require anything beyond existence.


docs/rules/require-meta-docs-recommended.md
```diff
@@ -0,0 +1,39 @@
...
+  meta: { recommended: true },
+  create(context) {
+    /* ... */
+  },
+};
+```
+
+## Further Reading
+
+- [working-with-rules#options-schemas](https://eslint.org/docs/developer-guide/working-with-rules#options-schemas)
```
bmish, line 39:
```suggestion
- [Rule Structure](https://eslint.org/docs/latest/extend/custom-rules#rule-structure)
```

docs/rules/require-meta-docs-recommended.md
```diff
@@ -0,0 +1,39 @@
+# Require rules to implement a `meta.docs.recommended` property (`eslint-plugin/require-meta-docs-recommended`)
+
+<!-- end auto-generated rule header -->
+
+Defining a whether recommended value for each rule can help developers understand whether they're recommended.
```
bmish, line 5: I would expand on the pros and cons here to help developers understand if the rule makes sense for them:

```suggestion
Utilizing `meta.docs.recommended` makes it clear from each rule implementation whether a rule is part of the `recommended` config. Some plugins also have scripting for conveniently generating their config based on this flag.

However, this flag may not be appropriate for all plugins:

* Extra scripting/tooling is needed to keep the flags in sync with the config
* The flag may not scale to plugins that have multiple/many configs or don't have a recommended config
* Or some may simply prefer to keep the source of truth solely in the config rather than duplicating config membership data in rules
```

lib/utils.js
```diff
@@ -800,6 +800,25 @@ module.exports = {
     });
   },
 
+  getMetaProperty(propertyName, ruleInfo, scopeManager) {
```
bmish, line 803: Either this needs to be able handle any property e.g. `meta.schema` or it should be renamed `getMetaDocsProperty` since it only handles docs properties right now.

JoshuaKGoldberg:
D'oh, sorry! I completely missed the notification.

bmish:
Looks like there are still conflicts.

tests/lib/rules/require-meta-docs-recommended.js
```diff
@@ -0,0 +1,113 @@
...
+    `
+      module.exports = {
+        meta: { docs: { recommended: true } },
+        create(context) {}
+      };
+    `,
+    {
+      code: `
+        export default {
+          meta: { docs: { recommended: 'disallow unused variables' } },
```
bmish, line 23: Should we be requiring this property to be a boolean?

JoshuaKGoldberg: I would prefer not - in typescript-eslint we have it as a string or object.

JoshuaKGoldberg:
I'm finishing for the day, will fix the lint issue tomorrow.

tests/lib/rules/require-meta-docs-recommended.js
bmish: What are the use cases for that? I'm OK to leave as-is if you want, but personally, I've only ever seen it as a boolean.

bmish (approved):

tests/lib/rules/require-meta-docs-recommended.js
JoshuaKGoldberg: What would that look like on the typescript-eslint side? `meta.docs.config` with the `ConfigKey | Record<ConfigKey, any>` that have now?

JoshuaKGoldberg: Ah, I just found/remembered another config that uses a non-boolean: `eslint-plugin-vitest`. https://github.com/veritem/eslint-plugin-vitest/blob/288bfb5b27fdfd838daef91b974a2484006db9c9/src/rules/expect-expect.ts#L41. Though, I don't see this reflected anywhere - maybe this isn't relevant...

Anyway, what I'm pondering now is: it feels to me like `meta.docs.recommended` operates as a subset of what e.g. `meta.docs.config` has to offer. As in, `docs: { recommended: true }` is roughly the same as `docs: { config: "recommended" }`, just ever so slightly fewer characters. But `config` allows plugins to have multiple "flavors". Which is a more extensible option. But is a rename...

bmish: Yeah, I agree the `recommended` property can only represent a limited subset of potential information.

To represent rules that may belong to multiple configs or even have different severities in different configs, you probably need an object like:

```
configs: {
  recommend: true, // or perhaps a severity string like error/warn/off
  strict: ...
}
```

This is why I personally don't try to represent config information in the rules themselves except for plugins with a very simple config structure, and why I wanted to be sure we mentioned caveats to this rule like in https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/447#discussion_r1493366564

JoshuaKGoldberg: Thanks, this conversation has been helpful. I just added an opt-in `allowNonBoolean` option.

tests/lib/rules/require-meta-docs-recommended.js
```diff
@@ -0,0 +1,178 @@
...
+
+const rule = require('../../../lib/rules/require-meta-docs-recommended');
+const RuleTester = require('../eslint-rule-tester').RuleTester;
+
+const ruleTester = new RuleTester({
+  languageOptions: { sourceType: 'commonjs' },
+});
+
+ruleTester.run('require-meta-docs-recommended', rule, {
+  valid: [
```
bmish, line 11: Can you make sure we allow situations where we can't statically determine the value? Sometimes people may use functions or other dynamic values that we can't analyze, and we generally want to ensure that's allowed. We shouldn't only flag a problem when we detect an actual invalid value or the property is obviously missing.

```js
module.exports = {
        meta: { docs: { recommended: RECOMMENDED } },
        create(context) {}
      };
```

JoshuaKGoldberg: Added. I'll note that this and existing similar functions don't handle the case of `meta: (unknown)` or `meta: { docs: (unknown) }` well.

bmish (approved):

tests/lib/rules/require-meta-docs-recommended.js
aladdin-add: possible the option "allowedValues", it defaults to `[true, false]`;

The advantage of this is that when you configure ["strict"], if there is a spelling mistake(e.g. `"stricts"`) it will be caught.

JoshuaKGoldberg: That wouldn't be useful to us in typescript-eslint, since we have config objects.

aladdin-add (approved):
LGTM, thanks!

tests/lib/rules/require-meta-docs-recommended.js
bmish: `allowedValues` sounds like overkill to me. For most, it's a boolean, but if not, just let them specify whatever...

PR #449, opened 2024-02-20
feat: support named exports in ESM/TS

### LINKED ISSUE #375: Rules do not work with named exports
Hello, and thanks for the plugin.

I tried to use these rules on my plugin today, but none of them seem to work, meaning that none of the rules were triggered.
I think this is because this plugin only looks for code matching `export default foo` or `module.exports = foo`.

My plugin does not use these and instead uses named exports. My plugin is a modern ESLint plugin written in TypeScript, and [using default exports in TypeScript is considered an anti-pattern](https://basarat.gitbook.io/typescript/main-1/defaultisbad).

For example, here is what [one of my rules](https://github.com/IsaacScript/isaacscript/blob/main/packages/eslint-plugin-isaacscript/src/rules/no-explicit-array-loops.ts) looks like:

```ts
import { ESLintUtils } from "@typescript-eslint/utils";

const createRule = ESLintUtils.RuleCreator(
  (ruleName) =>
    `https://github.com/IsaacScript/isaacscript/blob/main/packages/eslint-plugin-isaacscript/docs/rules/${ruleName}.md`,
);

export type Options = [];
export type MessageIds = "noExplicitArray";

export const noExplicitArrayLoops = createRule<Options, MessageIds>({
  name: "no-explicit-array-loops",
  meta: {
    type: "problem",
    docs: {
      description: "Disallows explicit iteration for arrays",
      recommended: "recommended",
      requiresTypeChecking: true,
    },
    schema: [],
    messages: {
      noExplicitArray:
        'Explicit iteration over arrays is disallowed. (Get rid of the "Object.values()".)',
    },
    fixable: "code",
  },
  defaultOptions: [],
  create(context) {
    // Rule logic here.
  },
});
```

Thus, I propose that the "detect rule" logic of this plugin is updated in some way. If you can advise on what the correct solution is, I can submit a pull request, as I would love to use the rules that this project offers. Thank you.

43081j:
This adds support for named exports in ESM (whether it be TS or not).

For example:

```ts
export const rule: RuleModule = {
  create: () => { ... };
};
```

Also, exported symbols:

```ts
const rule = { ... };

export {rule};
```

While ESLint plugins are still usually CJS at time of writing this, many are written as ESM sources, and an increasing number are using named exports.

### Reviewer notes

You'll want to hide whitespace changes when reviewing this, as prettier reformatted a bunch of it thanks to changing the level of nesting/chaining.

Fixes #375

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -122,14 +122,18 @@ function getRuleExportsESM(ast, scopeManager) {
 
   for (const statement of ast.body) {
     switch (statement.type) {
+      // export default rule;
       case 'ExportDefaultDeclaration': {
         possibleNodes.push(statement.declaration);
         break;
       }
+      // export = rule;
       case 'TSExportAssignment': {
         possibleNodes.push(statement.expression);
         break;
       }
+      // export const rule = { ... };
+      // or export {rule};
       case 'ExportNamedDeclaration': {
         for (const specifier of statement.specifiers) {
           possibleNodes.push(specifier.local);
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -118,55 +118,78 @@ function isTypeScriptRuleHelper(node) {
  * Helper for `getRuleInfo`. Handles ESM and TypeScript rules.
  */
 function getRuleExportsESM(ast, scopeManager) {
-  return ast.body
-    .filter((statement) =>
-      [
-        'ExportDefaultDeclaration', // export default rule;
-        'TSExportAssignment', // export = rule;
-      ].includes(statement.type)
-    )
-    .map((statement) => statement.declaration || statement.expression)
+  const possibleNodes = [];
 
-    .reduce((currentExports, node) => {
-      if (node.type === 'ObjectExpression') {
-        // Check `export default { create() {}, meta: {} }`
-        return collectInterestingProperties(
-          node.properties,
-          INTERESTING_RULE_KEYS
-        );
-      } else if (isFunctionRule(node)) {
-        // Check `export default function(context) { return { ... }; }`
-        return { create: node, meta: null, isNewStyle: false };
-      } else if (isTypeScriptRuleHelper(node)) {
-        // Check `export default someTypeScriptHelper({ create() {}, meta: {} });
-        return collectInterestingProperties(
-          node.arguments[0].properties,
-          INTERESTING_RULE_KEYS
-        );
-      } else if (node.type === 'Identifier') {
-        // Rule could be stored in a variable before being exported.
-        const possibleRule = findVariableValue(node, scopeManager);
-        if (possibleRule) {
-          if (possibleRule.type === 'ObjectExpression') {
-            // Check `const possibleRule = { ... }; export default possibleRule;
-            return collectInterestingProperties(
-              possibleRule.properties,
-              INTERESTING_RULE_KEYS
-            );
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -118,55 +118,78 @@ function isTypeScriptRuleHelper(node) {
  * Helper for `getRuleInfo`. Handles ESM and TypeScript rules.
  */
 function getRuleExportsESM(ast, scopeManager) {
-  return ast.body
-    .filter((statement) =>
-      [
-        'ExportDefaultDeclaration', // export default rule;
-        'TSExportAssignment', // export = rule;
```
bmish, line 125: Can you keep these comments showing an example of each type?

lib/utils.js
bmish, line 1: Can you speak to the possibility of this causing additional false positives as I mentioned in https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/375#issuecomment-1694523294? Do you anticipate any? It may be worth testing on a few top ESLint plugins for a change like this. Is there anything else we should do to mitigate this from detecting named exports that aren't ESLint rules?

tests/lib/utils.js
```diff
@@ -65,15 +65,11 @@ describe('utils', () => {
     describe('the file does not have a valid rule (ESM)', () => {
```
bmish, line 65: Can you make sure there are test cases covering the named export behavior for CJS? It sounds like the behavior shouldn't change for that.

43081j:
Thanks for the quick review @bmish , I'll sort the changes tomorrow 👍

lib/utils.js
aladdin-add: According to the `eslint-remote-tester`'s result, no false positives were found. We can start with this implementation and improve it if we receive feedback. thoughts?

https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/7980943358/job/21791683343?pr=449

aladdin-add: A very primitive idea is allowing configuring the esquery selectors for the possible rules:

```js
{
  settings: {
    // to allow "export const rule = ..."
    "eslint-plugin": {"rule-selectors": ["ExportNamedDeclaration[declaration.type="VariableDeclaration"][declaration.declarations.0.id.name="rule"]"]}
  }
}
```

bmish: @aladdin-add is `eslint-remote-tester` actually showing us new lint violations (which could be false positives), or just crashes? It says `Results: No errors` but I'd be kind of surprised if there really are no lint violations in any of those repositories. We may also want to add more ESM plugins to the list it tests against.

43081j:
FYI i have at least resolved the two changes

not sure what you want to do about the testing. since it seems like @aladdin-add already ran the remote tester successfully

lib/utils.js
bmish: @43081j can you manually test this on a few larger/top ESM plugins? I don't think eslint-remote-tester, which runs during CI, detects false positives for us, only crashes.

43081j: i'll have a look into it 👍

43081j:
@bmish i ran it against:

- eslint-plugin-compat
- eslint-plugin-jest
- TSESLint
- eslint-plugin-vue
- eslint-plugin-prettier

and all seemed well

bmish (approved):
Thanks!

PR #453, opened 2024-03-28
feat: re-export flat configs "flat/*"

aladdin-add:
long-time plan:

v6: no action
v7: rename `flat/*` => `*`; eslintrc config `*` => `legacy/*`?
v8~: remove eslintrc supports


bmish:
I think the plan you've described makes sense and aligns with the recommendation I wrote in https://github.com/eslint/eslint/issues/18095#issuecomment-1951414598.

We could also get rid of the direct string imports like `require('eslint-plugin-eslint-plugin/configs/recommended')` in the next major version.

bmish (approved):

aladdin-add:
> We could also get rid of the direct string imports like require('eslint-plugin-eslint-plugin/configs/recommended') in the next major version.

I think it's fine to keep it, as long as it doesn't come with large maintenance costs. :)

bmish (approved):

bmish:
> > We could also get rid of the direct string imports like require('eslint-plugin-eslint-plugin/configs/recommended') in the next major version.
> 
> I think it's fine to keep it, as long as it doesn't come with large maintenance costs. :)

It doesn't have much maintenance cost, but long-term, I still think it's preferable to standardize on a single means of using flat config with the standard `configs` object.

PR #474, opened 2024-04-30
feat: eslint-utils => @eslint-community/eslint-utils

aladdin-add:
eslint-plugin-eslint-comments => @eslint-community/eslint-plugin-eslint-comments

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -15,15 +15,15 @@ const compat = new FlatCompat({
 module.exports = [
   ...compat.extends(
     'not-an-aardvark/node',
-    'plugin:eslint-comments/recommended',
+    'plugin:@eslint-community/eslint-comments/recommended',
     'plugin:prettier/recommended',
     'plugin:unicorn/recommended',
   ),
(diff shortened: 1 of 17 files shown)


voxpelli:
Would be great to get a linting into these projects to ensure we make these migrations for all of the community modules :P

aladdin-add:
That's why I don't want to rename - esp. a widely used package. :)

voxpelli:
@aladdin-add We could make a shared config that adds [no-restricted-require](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-restricted-require.md) / [no-restricted-import](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-restricted-import.md) for the old names and use that within all of our repositories

aladdin-add:
It can be done, but I'm not very sure if it's necessary. These migrations are all one-time, and once they're done, they're no longer needed. 😄

voxpelli:
Until a new module is transferred and we need to be reminded to do it all again ;P

bmish (approved):

PR #484, opened 2024-06-23
feat: [no-property-in-node] add additionalNodeTypeFiles option

### LINKED ISSUE #482: Bug: no-property-in-node doesn't recognize locally linked typescript-eslint types
Coming over from https://github.com/typescript-eslint/typescript-eslint/issues/8226 -> https://github.com/typescript-eslint/typescript-eslint/pull/9261: amusingly+ironically, the `eslint-plugin/no-property-in-node` rules doesn't detect any types in the https://github.com/typescript-eslint/typescript-eslint codebase. This is because typescript-eslint's types are defined at a path like:

```plaintext
/Users/josh/repos/typescript-eslint/packages/types/dist/generated/ast-spec.d.ts
```

...while the rule only checks `node_modules/` style paths:

https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/04e397b6d3c02084a68f5339f182fb5aecaaca5b/lib/rules/no-property-in-node.js#L3-L6

I suppose we have two possible fixes:

* Add a third path like `/packages[/\\]types[/\\]dist[/\\]generated[/\\]ast-spec\.d\.ts/,` to hardcode for typescript-eslint
* Add an option to allow user configs to specify their own paths

Proposal: can we go for the latter fix, adding a rule option? I feel awkward asking to hardcode this rule for one project. 😄

JoshuaKGoldberg:
Adds a new `additionalNodeTypeFiles` option as an array of regular expressions. User configs can specify additional files such as a path to a `node_modules/ ... .d.ts` file.

Fixes #482.

### DIFF
--- a/lib/rules/no-property-in-node.js
+++ b/lib/rules/no-property-in-node.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const typedNodeSourceFileTesters = [
+const defaultTypedNodeSourceFileTesters = [
   /@types[/\\]estree[/\\]index\.d\.ts/,
   /@typescript-eslint[/\\]types[/\\]dist[/\\]generated[/\\]ast-spec\.d\.ts/,
 ];
@@ -26,9 +26,10 @@ const typedNodeSourceFileTesters = [
  * ```
  *
  * @param {import('typescript').Type} type
+ * @param {RegExp[]} typedNodeSourceFileTesters
  * @returns Whether the type seems to include a known ESTree or TSESTree AST node.
  */
-function isAstNodeType(type) {
+function isAstNodeType(type, typedNodeSourceFileTesters) {
   return (type.types || [type])
     .filter((typePart) => typePart.getProperty('type'))
     .flatMap(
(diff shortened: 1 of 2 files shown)


lib/rules/no-property-in-node.js
```diff
@@ -55,13 +56,34 @@ module.exports = {
...
     },
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          additionalNodeTypeFiles: {
+            description:
+              'Any additional regular expressions to consider source files defining AST Node types.',
+            default: '^(enforce|require|disallow)',
```
aladdin-add, line 66: the `default` was copied from https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/04e397b6d3c02084a68f5339f182fb5aecaaca5b/lib/rules/require-meta-docs-description.js#L32? 😄

JoshuaKGoldberg: Haha you got me

aladdin-add:
The docs also need to be updated; I'll take care of it in another PR.

Issue #310, opened 2022-10-07
Add TypeScript type-checking to all rules


bmish:
We don't need to convert our JavaScript files to TypeScript, but by adding JSDoc comments with types and using the TypeScript [checkJs](https://www.typescriptlang.org/tsconfig#checkJs) option, we can get the benefits of type-checking. See this PR for an example and list of benefits: https://github.com/square/eslint-plugin-square/pull/723.

We also want to export types for consumers.

bmish:
I'm open to that as a full migration to TypeScript is the end-goal. Jsdoc comments can just be a nice intermediary step to get there.

We would want to do this as a follow-up to the ESM migration which is already open:
- #516

michaelfaith:
Great, i'd be down to help with that.

bmish:
Here's another example of the intermediary jsdoc type-annotation conversion (even just this can turn into a huge PR):

- https://github.com/qunitjs/eslint-plugin-qunit/pull/571

michaelfaith:
Right on.  If anyone wants to take a stab at that interim solution, feel free.  

I've started on the full TypeScript migration here: https://github.com/eslint-community/eslint-plugin-eslint-plugin/compare/main...michaelfaith:eslint-plugin-eslint-plugin:build/typescript  I'm going rule by rule, and then will do the tests one-by-one.  I've partially migrated the big `utils` module, but I'm just doing that as I migrate rules that use the functions.  

I'll be going out of town next week, and so I won't make a ton more progress until I'm back.

michaelfaith:
Supporting PR, moving the repo's eslint config to TypeScript (which is necessary to continue using the plugin locally once it's in typescript, without needing to build first) https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/531

michaelfaith:
Another supporting PR, removing the (now unnecessary) context compatibility functions.  I should have removed those when dropping support for ESLint v8.  And I did it in the TS branch, but that already is going to have a ton of changes.  So, breaking this out separately.  https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/532

michaelfaith:
I wrapped up the primary migration, and just working on adjusting the build / ci scripts.  Here's another supporting change, moving the `eslint-remote-tester` config to typescript: https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/533

[...]

PR #488, opened 2024-08-01
feat: add type comment to exported plugin

### LINKED ISSUE #487: Incorrect type inference for exported configs object: {}
Right now, when I try to use the plugin in files with type-aware linting and typescript-eslint's recommendedTypeChecked config, I get:

```plaintext
/Users/josh/repos/typescript-eslint-examples/packages/eslint-plugin-example-typed-linting/eslint.config.mjs
  9:5  error  Unsafe argument of type `any` assigned to a parameter of type `ConfigWithExtends`  @typescript-eslint/no-unsafe-argument
```

...on lines like:

```ts

export default tseslint.config(
    eslintPlugin.configs['flat/recommended'],
);
```

Example: https://github.com/typescript-eslint/examples/blob/b91625d418f63b4bcc5127f2fb213222dcadf3cc/packages/eslint-plugin-example-typed-linting/eslint.config.mjs

The root cause is that the module's exported `config` object is inferred from `.js` to be `{}`:

https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/95b1980c87ba1a35a1c9e3e45a3f68b170a30bf4/lib/index.js#L51

Normally for a project that isn't written in TypeScript I'd just send a PR to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped). But because this project already has #310 in its issues, I figured I'd give this a shot. 🙂

JoshuaKGoldberg:
Fixes #487.

This will give a bit nicer of an IDE experience for folks working in editors such as VS Code that have built-in TypeScript support.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -42,6 +42,7 @@ const allRules = Object.fromEntries(
     ]),
 );
 
+/** @type {import("eslint").ESLint.Plugin} */
 const plugin = {
   meta: {
     name: packageMetadata.name,


voxpelli:
Shouldn't help as TS doesn't read JSDoc comments from `node_modules`?

Would need to compile it to a .d.ts?

JoshuaKGoldberg:
`parserOptions.projectService` picks it up, it happens in the repro linked in the issue.

voxpelli:
Even when `maxNodeModuleJsDepth` hasn’t been set to something other than the default `0`?

Such behavior has already caused issues in ESLint as TS has an undocumented behavior of setting it when using `jsconfig.json`: https://github.com/eslint/eslint/issues/18100#issuecomment-1971500684

JSDoc comments should not be relied upon

JoshuaKGoldberg:
> Even when `maxNodeModuleJsDepth` hasn’t been set to something other than the default `0`?

Yes, per the reproduction case I linked in the issue.

Agreed that `.d.ts` files would be better. That's more work that I don't have time for right now.

Personally, I'd rather lean towards _"perfect is the enemy of good"_ and ship this now + hope for that eventually. There's no harm in including these type-rich JSDocs for pure `.js` codebases. They benefit the codebase's development itself, help pave the path for more formal TS adoption, and can also help in cases like the reproduction I linked in the issue.

voxpelli:
> > Even when `maxNodeModuleJsDepth` hasn’t been set to something other than the default `0`?
> 
> Yes, per the reproduction case I linked in the issue

So typescript eslint is not respecting the tsconfig in this regard?

I’m pro fixing #310 but against making any changes to JSDoc comments because others have decided to consume them as published types. https://github.com/eslint/eslint/issues/18100#issuecomment-1971500684 is one of many proofs that it’s a bad idea to consume JSDoc comments from third party modules

Unless this module publishes type definitions this module should be interpreted as untyped is my opinion

bmish (approved):

voxpelli (changes requested):
I don't object to the change itself, but I object to it being a solution to the problem at hand

JoshuaKGoldberg:
@voxpelli it's also useful for internal development and doesn't impact users using sanctioned type methods. If we ignore the reasoning around external types, do you have any remaining blockers?

voxpelli:
@JoshuaKGoldberg no other blockers, if we change it so that this doesn't promote usage of it externally in any way then I'm okay with it

JoshuaKGoldberg:
The comment as-is is exactly what one would do for a purely internal change. There's nothing to be done to promote or not-promote it externally.

voxpelli:
Removing the Fixes 😜

bmish (approved):

JoshuaKGoldberg:
Thanks all!

PR #490, opened 2024-10-18
feat: add `require-meta-schema-description` rule

### LINKED ISSUE #334: Rule: require that properties in a schema include a description
Porting over from https://github.com/typescript-eslint/typescript-eslint/issues/5392: it's theoretically possible for docs generators to use the JSON schema `description` fields on rule options per https://json-schema.org/draft/2020-12/json-schema-core.html#name-schema-vocabularies. That way auto-generated rule docs can automatically be set up with those descriptions.

Is this something you'd be interested in taking in as a core rule here? I'd imagine it wouldn't be something to enable in the _recommended_ config to start given these descriptions aren't used by the default markdown docs generator. But it might be useful long-term as an opt-in option to enforce that the descriptions live in the page somehow?

---

For a concrete example, in typescript-eslint, the `no-explicit-any` rule defines descriptions:

```js
schema: [
  {
    type: 'object',
    additionalProperties: false,
    properties: {
      fixToUnknown: {
        description:
          'Whether to enable auto-fixing in which the `any` type is converted to the `unknown` type.',
        type: 'boolean',
      },
      ignoreRestArgs: {
        description: 'Whether to ignore rest parameter arrays.',
        type: 'boolean',
      },
    },
  },
],
```

...and then https://typescript-eslint.io/rules/no-explicit-any/#options includes them as JSDoc options:

```ts
interface Options {
  /**
   * Whether to enable auto-fixing in which the `any` type is converted to the `unknown` type.
   */
  fixToUnknown?: boolean;
  /**
   * Whether to ignore rest parameter arrays.
   */
  ignoreRestArgs?: boolean;
}

const defaultOptions: Options = [
  { fixToUnknown: false, ignoreRestArgs: false },
];
```

JoshuaKGoldberg:
Fixes #334.

This is a relatively straightforward approach that descends through the AST of `meta.schema`. It doesn't attempt to dynamically construct the properties through AJV. 

I tried it out on typescript-eslint here: https://github.com/typescript-eslint/typescript-eslint/pull/10171

### DIFF
--- a/lib/rules/require-meta-schema-description.js
+++ b/lib/rules/require-meta-schema-description.js
@@ -1,0 +1,102 @@
+'use strict';
+
+const utils = require('../utils');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'require rules `meta.schema` properties to include descriptions',
+      category: 'Rules',
+      recommended: true,
+      url: 'https://github.com/eslint-community/eslint-plugin-eslint-plugin/tree/HEAD/docs/rules/require-meta-schema-description.md',
+    },
+    schema: [],
+    messages: {
+      missingDescription: 'Schema option is missing an ajv description.',
+    },
+  },
+
+  create(context) {
+    const sourceCode = context.sourceCode || context.getSourceCode(); // TODO: just use context.sourceCode when dropping eslint < v9
+    const { scopeManager } = sourceCode;
+    const ruleInfo = utils.getRuleInfo(sourceCode);
+    if (!ruleInfo) {
+      return {};
+    }
+
+    const schemaNode = utils.getMetaSchemaNode(ruleInfo.meta, scopeManager);
+    if (!schemaNode) {
+      return {};
+    }
+
+    const schemaProperty = utils.getMetaSchemaNodeProperty(
+      schemaNode,
+      scopeManager,
+    );
+    if (schemaProperty?.type !== 'ArrayExpression') {
+      return {};
+    }
+
+    for (const element of schemaProperty.elements) {
+      checkSchemaElement(element, true);
+    }
+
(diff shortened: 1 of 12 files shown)


MichaelDeBoey (approved):

lib/rules/require-meta-schema-description.js
```diff
@@ -0,0 +1,102 @@
...
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'require rules `meta.schema` properties to include descriptions',
+      category: 'Rules',
+      recommended: true,
```
bradzacher, line 17: breaking change?

lib/rules/require-meta-schema-description.js
```diff
@@ -0,0 +1,102 @@
...
+    function checkSchemaElement(node, isRoot) {
+      if (node.type !== 'ObjectExpression') {
+        return;
+      }
+
+      let hadChildren = false;
+      let hadDescription = false;
+
+      for (const { key, value } of node.properties) {
+        switch (key?.name) {
```
bradzacher, line 62: (1) we shouldn't match
```js
const description = 'lol';
({ [description]: ... });
```

(2) we can support quoted keys
```js
{ 'description': ... }
```

```suggestion
        if (key.computed) {
          continue;
        }
        switch (key.name ?? key.value) {
```

bradzacher (approved):

lib/rules/require-meta-schema-description.js
aladdin-add: yes, it follows the same policy as eslint: https://github.com/eslint/eslint?tab=readme-ov-file#semantic-versioning-policy.

aladdin-add:
@JoshuaKGoldberg the ci was failing somehow, can you look please?

JoshuaKGoldberg:
Ah, I was missing some test coverage in the latest changes. Filled it in & used the `getStaticValue` utility to be more robust around detecting keys.

aladdin-add (approved):
LGTM, thanks!

PR #495, opened 2024-11-17
feat!: capitalize default pattern of `require-meta-docs-description` rule

azat-io:
This PR modifies the default pattern value of ESLint rules description.

Currently almost all ESLint plugins use descriptions starting with a capital letter.

ESLint rules, example:
https://github.com/eslint/eslint/blob/main/lib/rules/accessor-pairs.js#L149

TypeScript ESLint rules, example:
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/adjacent-overload-signatures.ts#L29

ESLint React:
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/boolean-prop-naming.js#L47

ESLint Import:
https://github.com/un-ts/eslint-plugin-import-x/blob/master/src/rules/consistent-type-specifier-style.ts#L57

ESLint Promise:
https://github.com/eslint-community/eslint-plugin-promise/blob/main/rules/always-return.js#L162

ESLint Unicorn:
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/better-regex.js#L138

ESLint Perfectionist:
https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/rules/sort-array-includes.ts#L104

I think we should change the default behavior of the rule, since even ESLint itself doesn't adhere to it and writes descriptions for rules with a capital letter.


bmish:
If most of the top plugins use sentence case, then I'd be open to this.

For context, I found the PR where ESLint switched over to sentence case descriptions:

* https://github.com/eslint/eslint/pull/16052

azat-io:
Rebased

azat-io:
@aladdin-add Are there any plans to release a new major?

Zamiell:
@aladdin-add friendly ping, just in case this dropped off your radar

Zamiell:
@bmish Would you be able to take a look?

azat-io:
Rebased

PR #503, opened 2024-11-18
Add new `no-meta-schema-default` rule

### LINKED ISSUE #498: Rule suggestion: `no-meta-schema-default`
https://github.com/eslint/eslint/pull/17656 added support for `meta.defaultOptions` in rules and [mentions in the docs](https://github.com/eslint/eslint/pull/17656/files#diff-dc3f44ef800795b890a57c8dd95c6ce26ba8a47a4551d7a05d7a5cd9cf016597R846-R848) that Ajv's `useDefaults` option might be disabled in the future. See also https://github.com/eslint/rfcs/blob/main/designs/2023-rule-options-defaults/README.md:

> This RFC proposes removing the `useDefaults: true` setting in a future major version of ESLint.

To prepare for that, a new ESLint rule could warn against using `default` in `meta.schema` already.

FloEdelmann:
Fixes #498.

### DIFF
--- a/lib/rules/no-meta-schema-default.js
+++ b/lib/rules/no-meta-schema-default.js
@@ -57,8 +57,8 @@ module.exports = {
         return;
       }
 
-      for (const { key, value } of node.properties) {
-        if (!key) {
+      for (const { type, key, value } of node.properties) {
+        if (type !== 'Property') {
           continue;
         }
         const staticKey =
--- a/lib/rules/no-meta-schema-default.js
+++ b/lib/rules/no-meta-schema-default.js
@@ -1,0 +1,113 @@
+'use strict';
+
+const { getStaticValue } = require('@eslint-community/eslint-utils');
+const utils = require('../utils');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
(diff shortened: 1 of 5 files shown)


lib/rules/no-meta-schema-default.js
```diff
@@ -0,0 +1,113 @@
...
+
+    function checkSchemaElement(node) {
+      if (node.type !== 'ObjectExpression') {
+        return;
+      }
+
+      for (const { key, value } of node.properties) {
+        if (!key) {
+          continue;
+        }
```
aladdin-add, line 63: nits: It might be clearer to compare "type" directly (I'm assuming that the intention here is to skip the `...obj`):

```suggestion
      for (const { type, key, value } of node.properties) {
        if (type !== 'Property') {
          continue;
        }
```

aladdin-add (approved):
LGTM. Waiting a few days before merging.

aladdin-add:
it has been shipped in [v6.4.0](https://github.com/eslint-community/eslint-plugin-eslint-plugin/releases/tag/v6.4.0) 🎉

PR #516, opened 2025-06-15
feat!: move to ESM only

43081j:
Anyone who is using commonjs on node `<20` will have to do one of the following: 
- move to esm
- move to`eslint.config.mjs`
- upgrade node

Everyone else should be unaffected. 

This would be a new major of course. 

Let me know if there's any other changes you think we should do

### DIFF
--- a/.eslint-doc-generatorrc.js
+++ b/.eslint-doc-generatorrc.js
@@ -1,7 +1,7 @@
 import prettier from 'prettier';
 
 /** @type {import('eslint-doc-generator').GenerateOptions} */
-export default {
+const config = {
   ignoreConfig: [
     'all',
     'all-type-checked',
@@ -27,3 +27,5 @@ export default {
   urlConfigs:
     'https://github.com/eslint-community/eslint-plugin-eslint-plugin#presets',
 };
+
+export default config;
--- a/.eslint-doc-generatorrc.js
+++ b/.eslint-doc-generatorrc.js
@@ -1,9 +1,7 @@
-'use strict';
-
-const prettier = require('prettier');
+import prettier from 'prettier';
 
 /** @type {import('eslint-doc-generator').GenerateOptions} */
-module.exports = {
+export default {
   ignoreConfig: [
     'all',
     'all-type-checked',
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ yarn.lock
 
 # eslint-remote-tester
 eslint-remote-tester-results
+
+coverage/
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,3 @@ yarn.lock
 
 # eslint-remote-tester
 eslint-remote-tester-results
-
-coverage/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 .idea
-.nyc_output
+coverage
 .vscode
 node_modules/
 npm-debug.log
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,15 +3,43 @@
  * @author Teddy Katz
  */
 
-'use strict';
-
 // ------------------------------------------------------------------------------
 // Requirements
 // ------------------------------------------------------------------------------
 
-const fs = require('fs');
-const path = require('path');
-const packageMetadata = require('../package');
+import packageMetadata from '../package.json' with { type: 'json' };
+import consistentOutput from './rules/consistent-output.js';
+import fixerReturn from './rules/fixer-return.js';
+import metaPropertyOrdering from './rules/meta-property-ordering.js';
+import noDeprecatedContextMethods from './rules/no-deprecated-context-methods.js';
+import noDeprecatedReportApi from './rules/no-deprecated-report-api.js';
+import noIdenticalTests from './rules/no-identical-tests.js';
+import noMetaSchemaDefault from './rules/no-meta-schema-default.js';
+import noMissingMessageIds from './rules/no-missing-message-ids.js';
+import noMissingPlaceholders from './rules/no-missing-placeholders.js';
+import noOnlyTests from './rules/no-only-tests.js';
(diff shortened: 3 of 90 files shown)


43081j:
One of the ci failures seems to be some funk with unsupported API imports

I'll figure it out tomorrow

lib/utils.js
```diff
@@ -225,12 +220,12 @@ function getRuleExportsCJS(ast, scopeManager) {
       ) {
         exportsVarOverridden = true;
         if (isFunctionRule(node.right)) {
-          // Check `module.exports = function (context) { return { ... }; }`
```
bmish: Looks like a mistake since we want to continue to support commonjs rules.

.eslint-doc-generatorrc.js
```diff
@@ -1,9 +1,7 @@
-'use strict';
-
-const prettier = require('prettier');
+import prettier from 'prettier';
 
 /** @type {import('eslint-doc-generator').GenerateOptions} */
-module.exports = {
+export default {
```
bmish: I believe we need to update this for the jsdoc comment to work properly:

```suggestion
const config = {
```

and below:

```js
export default config;
```

lib/index.js
bmish, line 1: This is generated by hand now? Authors of new rules usually forget to update this kind of thing so just want to confirm `tests/index.js` verifies everything is included here properly?

michaelfaith:
For the node 21-23 test failures, I wonder if it's just the typechecking taking slightly longer in that first run of the rule.  The earlier node versions that succeeded, the runtime of that rule was pretty long too, and somewhat close to the 2s timeout

image

lib/index.js
43081j: it is

we could write a script to auto generate it if that helps

bmish: It's fine for now.

lib/rules/no-property-in-node.js
bmish, line 1: Thread about the test failure in Node 21-24.

Looks like it's just this rule only.

```
  1068 passing (7s)
  1 failing

  1) no-property-in-node
       valid
         'a' in window;:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/eslint-plugin-eslint-plugin/eslint-plugin-eslint-plugin/tests/lib/rules/no-property-in-node.js)
      at process.processImmediate (node:internal/timers:478:21)
```

CC: @JoshuaKGoldberg who authored the rule in case he has any ideas

bmish: I'm updating TS dependencies just in case:
- https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/523

michaelfaith: Even if it's not something that gets merged, it'd be interesting to know if extending the timeout does indeed allow the tests to pass, or if it's something more significant

michaelfaith: I'll try and repro it locally, when i get home

michaelfaith: Running locally passes for me in half the timeout time (but still long).  This is on node 22.  So, that would suggest it's environment dependent and the type checking is running a bit longer on node 21-24.

image

michaelfaith: Looking at the builds in https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/523, the time for that rule didn't markedly change between 20 and 21.  It's consistently around 1s (https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/15759677786/job/44423115996)
image

So, there definitely seems to be a regression (possibly on the typescript-eslint side?).

bmish: I'm not seeing a regression. The time for this expensive test case is the same before and after your PR. `'a' in window;` is slow because the `window` object is so large.

So it seems like something about the test environment is failing on the slow test case now. Need to figure out why that is.

Also briefly looked at optimizing the rule performance which would be nice regardless but didn't come up with anything.

bmish: The PR switches from c8 to nyc as nyc doesn't appear to support esm. I'm guessing c8 is either causing the test to run slower or has a lower test timeout.

bmish: Increasing the mocha timeout from 2000ms to 3000ms (`--timeout 3000`) didn't fix it. Unclear how long the timeout would need to be to accommodate this test file.

michaelfaith: I could take a stab at moving the repo to `vitest` (in place of mocha / c8) in a separate change, if that would help.

bmish: Yes, would be open to a separate PR to move to vitest first, assuming that has advantages over the current setup.

michaelfaith: https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/525

michaelfaith: @43081j this merged, if you want to rebase when you're able.

.gitignore
```diff
@@ -8,3 +8,5 @@ yarn.lock
 
 # eslint-remote-tester
 eslint-remote-tester-results
+
+coverage/
```
bmish: This is already in the list above so we can remove:

```suggestion
```

lib/utils.js
```diff
@@ -343,685 +338,656 @@ function collectArrayElements(node) {
   return [];
 }
 
-module.exports = {
-  /**
-  * Performs static analysis on an AST to try to determine the final value of `module.exports`.
```
bmish, line 348: revert a few mistaken replacements in the comments here

bmish (approved):
Nice work!

PR #518, opened 2025-06-17
New rule: `no-meta-replaced-by`

### LINKED ISSUE #517: New Rule: `no-meta-replaced-by`
Based on the newer format for ESLint rule deprecation info:
- https://eslint.org/docs/latest/extend/rule-deprecation#-deprecatedinfo-type
- https://github.com/eslint/rfcs/tree/main/designs/2024-deprecated-rule-metadata

`meta.deprecated` should be used in place of `meta.replacedBy`. `meta.deprecated` has a more standardized format with many possible fields to fully represent a rule deprecation.

error-four-o-four:
Hey there,

This pull request introduces a new rule, `no-meta-replaced-by` to disallow the use of the `replacedBy` property at the top level of rule `meta`.

I'd appreciate any improvements to the non-generated sections in the docs.

Closes #517

### DIFF
--- a/docs/rules/no-meta-replaced-by.md
+++ b/docs/rules/no-meta-replaced-by.md
@@ -53,6 +53,10 @@ module.exports = {
 };
 ```
 
+## When Not To Use It
+
+If you do not plan to provide rule's documentation in website, you can turn off this rule.
+
 ## Further Reading
 
 - [ESLint docs: `DeprecatedInfo`](https://eslint.org/docs/latest/extend/rule-deprecation#-deprecatedinfo-type)
--- a/docs/rules/no-meta-replaced-by.md
+++ b/docs/rules/no-meta-replaced-by.md
@@ -1,0 +1,59 @@
+# Disallow rules to use a `meta.replacedBy` property (`eslint-plugin/no-meta-replaced-by`)
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/eslint-community/eslint-plugin-eslint-plugin#presets).
+
+<!-- end auto-generated rule header -->
+
+ESLint v9.21.0 introduces a new `DeprecatedInfo` type to describe whether a rule is deprecated and how it can be replaced. The legacy format used the meta property `replacedBy` which should be defined inside `deprecated` instead of at the top level.
+
+Examples of a correct usage can be found at [array-bracket-newline](https://github.com/eslint/eslint/blob/4112fd09531092e9651e9981205bcd603dc56acf/lib/rules/array-bracket-newline.js#L18-L38) and [typescript-eslint/no-empty-interface](https://github.com/typescript-eslint/typescript-eslint/blob/af94f163a1d6447a84c5571fff5e38e4c700edb9/packages/eslint-plugin/src/rules/no-empty-interface.ts#L19-L30)
+
+## Rule Details
+
+This rule disallows the `replacedBy` property in rules' `meta`.
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/no-meta-replaced-by: error */
+
+module.exports = {
+  meta: {
+    deprecated: true,
+    replacedBy: [],
+  },
+  create() {},
+};
+```
+
+Examples of **correct** code for this rule:
+
+```js
+/* eslint eslint-plugin/no-meta-replaced-by: error */
+
+module.exports = {
+  meta: {
+    deprecated: {
+      message: 'The new rule adds more functionality',
+      replacedBy: [
+        {
+          rule: {
+            name: 'the-new-rule',
+          },
+        },
+      ],
+    },
+  },
+  create() {},
+};
+
+module.exports = {
+  meta: {},
+  create() {},
+};
+```
+
+## Further Reading
+
+- [ESLint docs: `DeprecatedInfo`](https://eslint.org/docs/latest/extend/rule-deprecation#-deprecatedinfo-type)
+- [RFC introducing `DeprecatedInfo` type](https://github.com/eslint/rfcs/blob/main/designs/2023-rule-options-defaults/README.md)
--- a/lib/rules/no-meta-replaced-by.js
+++ b/lib/rules/no-meta-replaced-by.js
@@ -1,0 +1,63 @@
+/**
+ * @fileoverview Disallows the usage of `meta.replacedBy` property
+ */
+
+'use strict';
+
+const utils = require('../utils');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow rules to use a `meta.replacedBy` property',
+      category: 'Rules',
+      recommended: true,
+      url: 'https://github.com/eslint-community/eslint-plugin-eslint-plugin/tree/HEAD/docs/rules/no-meta-replaced-by.md',
+    },
+    schema: [],
+    messages: {
+      useNewFormat:
+        'Use `meta.deprecated.replacedBy` instead of `meta.replacedBy`',
+    },
+  },
+  create(context) {
+    const sourceCode = utils.getSourceCode(context);
+    const ruleInfo = utils.getRuleInfo(sourceCode);
+
+    if (!ruleInfo) {
+      return {};
+    }
+
+    return {
+      Program() {
+        const metaNode = ruleInfo.meta;
+
+        if (!metaNode) {
+          return;
+        }
+
+        const replacedByNode = utils
+          .evaluateObjectProperties(metaNode, sourceCode.scopeManager)
+          .find(
+            (p) =>
+              p.type === 'Property' && utils.getKeyName(p) === 'replacedBy',
+          );
+
+        if (!replacedByNode) {
+          return;
+        }
+
+        context.report({
+          node: replacedByNode,
+          messageId: 'useNewFormat',
+        });
+      },
+    };
+  },
+};
--- a/tests/lib/rules/no-meta-replaced-by.js
+++ b/tests/lib/rules/no-meta-replaced-by.js
@@ -1,0 +1,132 @@
+/**
+ * @fileoverview require rules to implement a `meta.type` property
+ * @author 唯然<weiran.zsd@outlook.com>
+ */
+
+'use strict';
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-meta-replaced-by');
+const RuleTester = require('../eslint-rule-tester').RuleTester;
+
+// ------------------------------------------------------------------------------
+// Tests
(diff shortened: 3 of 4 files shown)


docs/rules/no-meta-replaced-by.md
```diff
@@ -0,0 +1,63 @@
...
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/no-meta-replaced-by: error */
+
+module.exports = {
+  meta: {
+    deprecated: true,
+    replacedBy: [],
```
bmish:
```suggestion
    replacedBy: ['the-new-rule'],
```

lib/rules/no-meta-replaced-by.js
```diff
@@ -0,0 +1,63 @@
...
+// ------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow rules to use a `meta.replacedBy` property',
+      category: 'Rules',
+      recommended: true,
```
bmish: Making a rule `recommended` is a breaking change so we'll need to do this later.

```suggestion
      recommended: false,
```

tests/lib/rules/no-meta-replaced-by.js
```diff
@@ -0,0 +1,132 @@
...
+// ------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-meta-replaced-by');
+const RuleTester = require('../eslint-rule-tester').RuleTester;
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const valid = [
```
bmish, line 18: Can you add a test case indicating that `deprecated: true` is valid.

docs/rules/no-meta-replaced-by.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## When Not To Use It
+
+If you do not plan to provide rule's documentation in website, you can turn off this rule.
+
+## Further Reading
+
+- [ESLint docs: `DeprecatedInfo`](https://eslint.org/docs/latest/extend/rule-deprecation#-deprecatedinfo-type)
+- [RFC introducing `DeprecatedInfo` type](https://github.com/eslint/rfcs/blob/main/designs/2023-rule-options-defaults/README.md)
```
bmish: Wrong link

docs/rules/no-meta-replaced-by.md
```diff
@@ -0,0 +1,63 @@
...
+
+<!-- end auto-generated rule header -->
+
+ESLint v9.21.0 introduces a new `DeprecatedInfo` type to describe whether a rule is deprecated and how it can be replaced. The legacy format used the meta property `replacedBy` which should be defined inside `deprecated` instead of at the top level.
+
+Examples of a correct usage can be found at [array-bracket-newline](https://github.com/eslint/eslint/blob/4112fd09531092e9651e9981205bcd603dc56acf/lib/rules/array-bracket-newline.js#L18-L38) and [typescript-eslint/no-empty-interface](https://github.com/typescript-eslint/typescript-eslint/blob/af94f163a1d6447a84c5571fff5e38e4c700edb9/packages/eslint-plugin/src/rules/no-empty-interface.ts#L19-L30)
+
+## Rule Details
+
+This rule disallows the `replacedBy` property in rules' `meta`.
```
bmish:
```suggestion
This rule disallows the `meta.replacedBy` property in a rule.
```

lib/rules/no-meta-replaced-by.js
```diff
@@ -0,0 +1,63 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow rules to use a `meta.replacedBy` property',
```
bmish:
```suggestion
      description: 'disallow using the `meta.replacedBy` rule property',
```

docs/rules/no-meta-replaced-by.md
```diff
@@ -0,0 +1,63 @@
+# Disallow rules to use a `meta.replacedBy` property (`eslint-plugin/no-meta-replaced-by`)
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/eslint-community/eslint-plugin-eslint-plugin#presets).
+
+<!-- end auto-generated rule header -->
+
+ESLint v9.21.0 introduces a new `DeprecatedInfo` type to describe whether a rule is deprecated and how it can be replaced. The legacy format used the meta property `replacedBy` which should be defined inside `deprecated` instead of at the top level.
```
bmish:
```suggestion
As of ESLint v9.21.0, the rule property `meta.deprecated` can be either a boolean or an object of type `DeprecatedInfo`. The `DeprecatedInfo` includes an optional `replacedBy` array that replaces the now-deprecated `meta.replacedBy` property.
```

docs/rules/no-meta-replaced-by.md
```diff
@@ -0,0 +1,63 @@
...
+
+Examples of a correct usage can be found at [array-bracket-newline](https://github.com/eslint/eslint/blob/4112fd09531092e9651e9981205bcd603dc56acf/lib/rules/array-bracket-newline.js#L18-L38) and [typescript-eslint/no-empty-interface](https://github.com/typescript-eslint/typescript-eslint/blob/af94f163a1d6447a84c5571fff5e38e4c700edb9/packages/eslint-plugin/src/rules/no-empty-interface.ts#L19-L30)
```
bmish:
```suggestion
Examples of correct usage:

- [array-bracket-newline](https://github.com/eslint/eslint/blob/4112fd09531092e9651e9981205bcd603dc56acf/lib/rules/array-bracket-newline.js#L18-L38)
- [typescript-eslint/no-empty-interface](https://github.com/typescript-eslint/typescript-eslint/blob/af94f163a1d6447a84c5571fff5e38e4c700edb9/packages/eslint-plugin/src/rules/no-empty-interface.ts#L19-L30)
```

docs/rules/no-meta-replaced-by.md
```diff
@@ -0,0 +1,64 @@
...
+module.exports = {
+  meta: {},
+  create() {},
+};
+```
+
+## When Not To Use It
+
+If you do not plan to provide rule's documentation in website, you can turn off this rule.
+
```
bmish:
```suggestion
```

docs/rules/no-meta-replaced-by.md
```diff
@@ -0,0 +1,64 @@
+# Disallow using the `meta.replacedBy` rule property (`eslint-plugin/no-meta-replaced-by`)
+
+<!-- end auto-generated rule header -->
+
+As of ESLint v9.21.0, the rule property `meta.deprecated` can be either a boolean or an object of type `DeprecatedInfo`. The `DeprecatedInfo` includes an optional `replacedBy` array that replaces the now-deprecated `meta.replacedBy` property.
```
bmish:
```suggestion
As of ESLint v9.21.0, the rule property `meta.deprecated` can be either a boolean or an object of type `DeprecatedInfo`. The `DeprecatedInfo` type includes an optional `replacedBy` array that replaces the now-deprecated `meta.replacedBy` property.
```

error-four-o-four: Err ... github displays two suggested changes. One with and another without `type` in the second sentence. Which one is correct?

error-four-o-four (commented):
Would it be appropriate to write 
```
## When Not To Use It

If you need to support versions of ESLint prior to v9.
```

lib/rules/no-meta-replaced-by.js
```diff
@@ -0,0 +1,63 @@
...
+
+const utils = require('../utils');
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
```
bmish, line 15: We could also add an autofixer to this rule to move `meta.replacedBy` to `meta.deprecated.replacedBy`. A follow-up to do that would be fine since it could be tricky to get it right.

error-four-o-four: Yes, thought so too. But I'd have to get down to the nitty-gritty of eslint and ASTs. The autofixer should be added in a follow-up.

docs/rules/no-meta-replaced-by.md
bmish: fixed, want to add `type`.

lib/rules/no-meta-replaced-by.js
bmish: Sounds good.

docs/rules/no-meta-replaced-by.md
bmish, line 1: > Would it be appropriate to write
> 
> ```
> ## When Not To Use It
> 
> If you need to support versions of ESLint prior to v9.
> ```

Regarding when not to use it, I don't think we necessarily need to include this section. In fact, anyone can use this rule regardless of ESLint version. As far as I know, ESLint itself does not validate nor use the `meta.deprecated` property. So anybody can use the object format for `meta.deprecated`.

Note that if someone is using TypeScript to define their rules, they would want to make sure they have up-to-date rule types (which I believe come from @types/eslint or typescript-eslint). And rule authors would want to make sure they are using up-to-date versions of any third-party tooling like [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) that use this property.

OR, you could include the section and just state the context I've mentioned here.

lib/rules/no-meta-replaced-by.js
error-four-o-four: Another fix in a follow-up might be to remove `replacedBy` from the `defaultOrder` in `meta-property-order`, right?

error-four-o-four: I'm also thinking about the neccessity and possibilities so that the autofixer is able to add the property `plugin` to the replacement rule according to `ReplacedByInfo`

If this is not the desired behaviour of the autofixer it would simply convert the items of the old `replacedBy` array to the new format e.g. `[{ rule: { name: 'the-new-rule' } }]`

If it is the desired behaviour one would have to determine and compare the prefix of the plugin. This could be achieved by either providing an option in `no-meta-replaced-by` or reading the value of an exported flat config which might be trickier and not reliable.

bmish: > Another fix in a follow-up might be to remove `replacedBy` from the `defaultOrder` in `meta-property-order`, right?

I think we can leave it. Doesn't hurt to continue to support the older rule properties.

bmish: Regarding the autofixer, it would be nice if the autofixer can work without any configuration. But if there's an option that can be provided to the rule to improve the autofixer further, I'm open to that as an added bonus.

- https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/519

bmish (approved):
Nice work!

error-four-o-four:
thank you @bmish for the guidance

PR #522, opened 2025-06-19
ci: add node v24 to test matrix

### LINKED ISSUE #521: Tooling: node v24 validation
Since node 24 is lined up to be the next LTS version in October, would it be worth adding v24 to the ci matrix for testing?

michaelfaith:
This change adds node v24 to the matrix for ci testing

Closes #521


bmish (approved):
Thanks!

PR #525, opened 2025-06-19
test: change test runner to vitest

michaelfaith:
This change moves the test runner from `mocha` + `nyc` to `vitest`.

In order to import `vitest` the tests had to be in esm, so I checked out everything under the `tests` folder from https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/516 (leaving the source as it was).

I also moved the `eslint-rule-tester` out of the `tests/lib` folder and into its own `utils` folder.  That way vitest didn't treat it as a test.

Note: rather than using `vitest`'s preferred `v8` coverage reporter, I used the `istanbul` one, to ensure the coverages between old and new were the same.  I did notice that when I tried the `v8` coverage reporter, the coverage numbers were much less.  Something to consider as a follow-up change.  The v8 reporter *should* be more accurate, so the coverage may not be as high as it seems.  I'd recommend moving up to the `v8` reporter after the esm branch lands.


package.json
```diff
@@ -60,7 +54,7 @@
     "@types/estree": "^1.0.8",
     "@typescript-eslint/parser": "^8.34.1",
     "@typescript-eslint/utils": "^8.34.1",
-    "chai": "^4.5.0",
+    "@vitest/coverage-istanbul": "^3.2.4",
```
bmish, line 57: sort dependencies

tests/lib/fixtures/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
     "moduleResolution": "NodeNext"
-  }
+  },
+  "files": ["estree.ts", "file.ts"]
```
bmish: Does this work?

```suggestion
  "files": ["*.ts"]
```

michaelfaith: Yeah, it'll be `include` instead of `files`, but I can change it to a wildcard

michaelfaith: Updated

package.json
michaelfaith: I believe that is correctly sorted.  Which is out of order?

bmish: nevermind, misread

package.json
```diff
@@ -107,5 +100,6 @@
     "npm": {
       "skipChecks": true
     }
-  }
+  },
+  "packageManager": "pnpm@10.12.1+sha512.f0dda8580f0ee9481c5c79a1d927b9164f2c478e90992ad268bbb2465a736984391d6333d2c327913578b2804af33474ca554ba29c04a8b13060a717675ae3ac"
```
bmish: Can we remove this? Our repo is currently using npm (in github scripts etc). If we want to switch to pnpm, we can do it in a separate PR.

eslint.config.js
```diff
@@ -2,7 +2,6 @@
 
 const js = require('@eslint/js');
 const { FlatCompat } = require('@eslint/eslintrc');
-const globals = require('globals');
```
bmish, line 5: Can we remove `globals` from package.json now?

package.json
michaelfaith: Yes, apologies, I didn't mean for that to go in.

michaelfaith: Updated

eslint.config.js
michaelfaith: Removed

Issue #505, opened 2024-11-20
V7 Planned Changes


aladdin-add:
Starting a list of my desired breaking changes, feedback welcome. We can perform this release once we have compiled enough changes.

Planned:
* Enable additional `recommended` rules:
   * https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/530
   * Note: the new rule `require-meta-docs-recommended` is NOT intended to become a `recommended` rule, so we won't enable this one.
* Change default rule options
   * There don't appear to be any changes needed to default rule options this time.
* Supported Node.js versions
  * https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/529
* Supported ESLint versions
  * https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/528
* Migrations
  * https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/516 (more info: [ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm))
  * https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/534

For reference, V6 was released 2024-04-10 ([CHANGELOG](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/releases/tag/v6.0.0), discussion in #269).

azat-io:
+1 for TS config support 👍

michaelfaith:
Need any help with this work?

bmish:
Nice, yes I'd be interested in moving to both ESM and TypeScript (separately is fine). Would there be any user-facing impact from the ESM conversion?

michaelfaith:
https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/528

43081j:
@bmish tried it out in TSESLint and all looks good!

works well 🎉

PR #528, opened 2025-06-20
feat!: remove eslint v8 / eslintrc support and remove `flat/` prefix from configs

michaelfaith:
This change removes support for legacy rc-based configs, and moves minimum supported version to 9.0.  I've also removed the deprecated `/configs` entry point.  Please use the plugin's `configs` object.

Contributes to https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/505


bmish:
Can you cherry-pick my commit from this PR to include the RuleTester update? Then I can close my PR.

- https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/524

michaelfaith:
Your PR removed our `RuleTester`, but with the `vitest` change, we still need to attach `vitest`'s functions to `RuleTester`.  So I removed the v8 specific logic from it, but left the `vitest` augmentation.

michaelfaith:
Oh, nice, I like your solution more. Thanks

tests/utils/test-setup.js
bmish, line 1: > Your PR removed our RuleTester, but with the vitest change (which I believe was still in PR, when your change went up), we still need to attach vitest's functions to RuleTester. So I removed the v8 specific logic from it, but left the vitest augmentation. I think this will cover what was in your change?

@michaelfaith I just pushed this change to solve that so rule test files can use the standard RuleTester import. Look good?

michaelfaith: Yep!  That's much cleaner. Thanks.

lib/index.js
bmish, line 1: Here's some more context about the config names btw:

- https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/453

I think the current plan to just drop ESLint v8 / eslintrc now is fine.

tests/lib/rules/require-meta-docs-url.js
bmish, line 1: There's a `.eslintrc.js` code sample in the rule doc for this rule. Can you update that to the new style?

README.md
```diff
@@ -42,25 +40,13 @@ Here's an example ESLint configuration that:
...
-  "rules": {
-    "eslint-plugin/require-meta-docs-description": "error"
-  }
-}
-```
-
-### <a name='flat'></a>[`eslint.config.js`](https://eslint.org/docs/latest/use/configure/configuration-files-new) (requires eslint>=v8.23.0)
+Note: you might need to set `sourceType` to `module` or `script` depending on your codebase.
 
 ```js
```
bmish, line 45: For clarity, it would be nice to include the filename in at least one of these examples to people don't try to use it with the old eslintrc format:

```suggestion
```js
// eslint.config.js
```

docs/rules/require-meta-docs-url.md
```diff
@@ -96,29 +96,54 @@ module.exports = {
...
-  plugins: ['eslint-plugin'],
-  rules: {
-    'eslint-plugin/require-meta-docs-url': [
-      'error',
-      {
-        pattern: `path/to/v${version}/docs/rules/{{name}}.md`,
-      },
-    ],
+import eslintPlugin from 'eslint-plugin-eslint-plugin';
+import packageMetadata from './package.json' with { type: 'json' };
```
bmish, line 130: > Error:   130:29  error  Can't resolve './package.json' in '/home/runner/work/eslint-plugin-eslint-plugin/eslint-plugin-eslint-plugin/docs/rules/require-meta-docs-url.md'  n/no-missing-import

You can disable that lint rule for this file if necessary.

bmish: Example of disabling a lint rule for markdown code sample:

https://github.com/ember-cli/eslint-plugin-ember/blob/ea0e3802b9971884eab5092e07730e4a0b04d9de/eslint.config.js#L172

michaelfaith: Ah, I had `'n/no-missing-imports'` instead of `'n/no-missing-import'`  doh

README.md
michaelfaith: Updated

tests/lib/rules/require-meta-docs-url.js
michaelfaith: Updated

docs/rules/require-meta-docs-url.md
michaelfaith: Looks good now

bmish (approved):
Thank you!

PR #529, opened 2025-06-21
feat!: remove support for EoL Node versions

michaelfaith:
This change removes support for Node v18, 21, and 23.  Supported versions are now `"^20.9.0 || ^22.11.0 || >=24.0.0"`

Contributes to #505


package.json
```diff
@@ -81,7 +81,7 @@
     "eslint": ">=9.0.0"
   },
   "engines": {
-    "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
+    "node": "^20.9.0 || ^22.11.0 || >=24.0.0"
```
bmish: The previous versions were based on ESLint:

https://github.com/eslint/eslint/blob/4ab44823df4d4b47d3650da949077a0551e7579e/package.json#L224

They usually choose high enough versions from each supported major version line in order to support new Node features they want to use. We can do the same.

Looking through the [changelogs](https://nodejs.org/en/about/previous-releases), how about we go with:
- https://github.com/nodejs/node/releases/tag/v22.13.1 - has notable features, about 6 months ago
- https://github.com/nodejs/node/releases/tag/v20.19.0 - has notable features, about 3-4 months ago

```suggestion
    "node": "^20.19.0 || ^22.13.1 || >=24.0.0"
```

PR #530, opened 2025-06-21
feat!: enable `no-meta-replaced-by`, `no-meta-schema-default`, `require-meta-default-options`, `require-meta-schema-description` as `recommended` rules

bmish:


bmish:
@aladdin-add can you look?

aladdin-add:
the rule `require-meta-docs-recommended` was also introduced in v6: https://github.com/eslint-community/eslint-plugin-eslint-plugin/releases/tag/v6.1.0

do we want to enable it too (possibly helpful for `eslint-doc-generator`)?

bmish:
I tried to make it clear originally in the [require-meta-docs-recommended rule doc](https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-docs-recommended.md) that this rule is not suitable to become a `recommended` rule:

> However, this rule may not be appropriate for all plugins:
> 
> - Extra scripting/tooling is needed to keep the flags in sync with the config
> - The flag may not scale to plugins that have multiple/many configs or don't have a recommended config
> - Or some may simply prefer to keep the source of truth solely in the config rather than duplicating config membership data in the rules

`eslint-doc-generator` doesn't rely on it.

So I've intentionally omitted it from this PR.

PR #531, opened 2025-07-19
build: migrate eslint.config.js to typescript 

michaelfaith:
This change migrates the eslint config for the repo to typescript.  This is in support of the larger migration to typescript.  In order for us to continue to consume the plugin locally, after it's been converted to typescript (without having to build first), we'll need to be loading the config from typescript.

Since I needed to update eslint to support, I took the opportunity to all of the other plugins too (except unicorn).

I also removed a rule disable that had a note on it about dropping after we dropped support for node 14.


tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "rootDir": ".",
+    "declaration": true,
+    "esModuleInterop": true,
+    "module": "esnext",
+    "moduleResolution": "bundler",
+    "noEmit": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "ES2022"
```
bmish:
```suggestion
    "target": "ES2024"
```

tsconfig.json
bmish, line 1: A few suggestions from a tsconfig.json I added recently for another plugin:

https://github.com/qunitjs/eslint-plugin-qunit/pull/571/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0

Can we add:

- `"verbatimModuleSyntax": true`,
- `"erasableSyntaxOnly": true`,  
- `"forceConsistentCasingInFileNames": true`,

michaelfaith: For sure.  I'll make those updates.  Question for you as well: when it comes to the TypeScript build in the other branch, would you prefer it to be bundled or not bundled?  I tend to default to bundled, which is what these settings reflect.  But if you prefer unbundled, then I'll change it to `nodenext` for `moduleResolution`.

bmish: If there's a common practice here by other ESLint plugins written/compiled in TypeScript, let's try to follow that. AI is telling me that most of them do not bundle, but I haven't dug into it too deeply.

michaelfaith: I did a quick check of the top 10 most popular plugins and configs (according to AI), and here are my findings:

| package                       | source     | bundled?  |
|-------------------------------|------------|-----------|
| @typescript-eslint/*          | TypeScript | unbundled |
| eslint-plugin-react           | JavaScript | N/A       |
| eslint-plugin-react-hooks     | TypeScript | bundled   |
| eslint-plugin/config-prettier | JavaScript | N/A       |
| eslint-plugin-import          | JavaScript | N/A       |
| eslint-plugin-jsx-a11y        | JavaScript | N/A |
| eslint-plugin-unicorn         | JavaScript | N/A       |
| eslint-plugin-jest            | TypeScript | unbundled |
| eslint-plugin-testing-library | TypeScript | unbundled |
| eslint-config-airbnb          | JavaScript | N/A       |
| eslint-plugin-storybook       | TypeScript | bundled   |

So, not a ton more, but leaning towards unbundled (from a sample size of 5 😅).  So, I'll go with that.

bmish: Is there any impact from this choice?

michaelfaith: Updated

tsconfig.json
michaelfaith: Updated

bmish (approved):

PR #533, opened 2025-07-23
build: convert eslint-remote-tester config to typescript

michaelfaith:
Supporting the larger typescript migration, this change migrates the config for `eslint-remote-tester` to typescript, which will be necessary to run our plugin from source, once it's typescript.

PR #534, opened 2025-07-27
feat: migrate package to TypeScript and publish types

### LINKED ISSUE #310: Add TypeScript type-checking to all rules

michaelfaith:
This change migrates all source (runtime code and tests) to TypeScript, and as a result now results in type definitions being published with the package.

Because I knew this was going to be a big PR, I tried to break things up the best I could by commit.  I first `git mv` all runtime files, and then in separate commits, migrated each rule.  I then did the same thing for tests.

The actual changes to test files, beyond the file name change, was fairly minimal.  Most of them were just simple file name changes.  Only a few needed some additional work (e.g. types added to utility functions).  So, I feel pretty good about maintaining functionality.  

I will say, the type in `utils` aren't the best, and I had to resort to adding `@ts-expect-error` in a couple places, but even then, they're in a pretty decent state.  I wanted to keep the flow of the functions as close to their original form as possible in this initial cut over.  All of that is to say, there's opportunity to iterate on these internal types, to improve them further.  Of course, those changes wouldn't be breaking and could potentially come after v7 is released.  

In addition to all the runtime and test file changes, I addded a build script (using `tsup`) and updated the `package.json` with new entry point paths and to include `dist` instead of `lib`.  I also updated the publish GH workflow to run a build before `npm publish`.  

Supporting work that I landed prior to this:
- migrate `eslint.config.js` to typescript so we could use the plugin directly from source: https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/531
- Remove unneeded context compatibility functions: https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/532
- migrate the `eslint-remote-tester` config to be typescript so we could use the plugin directly from source: https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/533

Closes #310


michaelfaith:
If there are any other slices of work you think should be peeled off and broken out as a separate change, let me know.  I _think_ everything in here, for the most part, need to go together, but open for input on that.

.gitignore
```diff
@@ -5,6 +5,7 @@ node_modules/
 npm-debug.log
 yarn.lock
 .eslintcache
+dist
```
bmish: nit: I like to use `dist/` to make it clear it's a directory.

```suggestion
dist/
```

lib/rules/fixer-return.ts
```diff
@@ -70,10 +83,9 @@ const rule = {
     /**
      * Check if a returned/yielded node is likely to be a fix or not.
      * A fix is an object created by fixer.replaceText() for example and returned by the fix function.
-     * @param {ASTNode} node - node to check
-     * @returns {boolean}
+     * @param node - node to check
```
bmish, line 85:
```suggestion
```

lib/rules/meta-property-ordering.ts
```diff
@@ -52,24 +54,29 @@ const rule = {
       return {};
     }
 
-    const order = context.options[0] || defaultOrder;
+    const order: string[] = context.options[0] || defaultOrder;
 
-    const orderMap = new Map(order.map((name, i) => [name, i]));
+    const orderMap = new Map<string | null, number>(
```
bmish, line 59: Technically, a Map can also return `undefined` when looking up a key that's not in it.

```suggestion
    const orderMap = new Map<string | null, number | undefined>(
```

If that's too much of a pain, you can ignore it for convenience, but that would be the ideal way to type it and then guard against a map lookup that doesn't return anything.

lib/rules/no-identical-tests.ts
```diff
@@ -27,20 +27,13 @@ const rule = {
...
     const sourceCode = context.sourceCode;
 
-    // ----------------------------------------------------------------------
-    // Helpers
-    // ----------------------------------------------------------------------
     /**
      * Create a unique cache key
-     * @param {object} test
-     * @returns {string}
+     * @param test
```
bmish:
```suggestion
```

lib/rules/no-meta-schema-default.ts
```diff
@@ -43,31 +43,32 @@ const rule = {
...
         }
+        const { key, value } = property;
         const staticKey =
           key.type === 'Identifier' ? { value: key.name } : getStaticValue(key);
         if (!staticKey?.value) {
           continue;
         }
 
-        switch (key.name ?? key.value) {
+        switch ('name' in key ? key.name : 'value' in key ? key.value : '') {
```
bmish: Normally I prefer to check / narrow based on node types so we know exactly what type we're dealing rather than using `in`. Does this work?
```suggestion
        switch (key.type === 'Identifier' ? key.name : key.value) {
```

lib/rules/no-missing-placeholders.ts
```diff
@@ -96,9 +96,9 @@ const rule = {
             messageId,
             data,
           } of reportMessagesAndDataArray.filter((obj) => obj.message)) {
-            const messageStaticValue = getStaticValue(message, scope);
+            const messageStaticValue = getStaticValue(message!, scope);
```
bmish: Why is `!` necessary? Let's prefer to use early return (`if (!message) {return}`) rather than manually changing the type.

lib/rules/no-only-tests.ts
```diff
@@ -53,9 +53,9 @@ const rule = {
                         const sourceCode = context.sourceCode;
 
                         const tokenBefore =
-                          sourceCode.getTokenBefore(onlyProperty);
+                          sourceCode.getTokenBefore(onlyProperty)!;
```
bmish: Let's prefer to use early return (`if (!onlyProperty) {return}`) rather than manually changing the type.

lib/rules/no-property-in-node.ts
```diff
@@ -23,11 +26,14 @@ const defaultTypedNodeSourceFileTesters = [
  * }
  * ```
  *
- * @param {import('typescript').Type} type
- * @param {RegExp[]} typedNodeSourceFileTesters
+ * @param type
+ * @param typedNodeSourceFileTesters
```
bmish: No need for jsdoc types anymore.
```suggestion
```

lib/rules/no-unused-message-ids.ts
```diff
@@ -99,21 +101,27 @@ const rule = {
...
+            .map((obj) => obj.messageId)
+            .filter((messageId) => !!messageId)) {
             const values =
               messageId.type === 'Literal'
                 ? [messageId]
-                : findPossibleVariableValues(messageId, scopeManager);
+                : findPossibleVariableValues(
+                    messageId as Identifier,
+                    scopeManager,
+                  );
```
bmish: Would prefer type checks over casting, and just bail out if it's not the right type.

```suggestion
                : typeof messageId === 'Identifier' ? findPossibleVariableValues(
                    messageId,
                    scopeManager,
                  ): [];
```

lib/rules/no-unused-message-ids.ts
```diff
@@ -99,21 +101,27 @@ const rule = {
...
               values.some((val) => val.type !== 'Literal')
             ) {
               // When a dynamic messageId is used and we can't detect its value, disable the rule to avoid false positives.
               hasSeenUnknownMessageId = true;
             }
-            values.forEach((val) => messageIdsUsed.add(val.value));
+            values.forEach(
+              (val) =>
+                'value' in val && messageIdsUsed.add(val.value as string),
+            );
```
bmish: It would be better to filter this properly by type without casting.
```suggestion
            values
            .filter(val=>val.type === 'Literal' && typeof val.value === 'string')
            .forEach(val => messageIdsUsed.add(val.value));
```

lib/rules/no-unused-message-ids.ts
```diff
@@ -127,18 +135,23 @@ const rule = {
...
+              : findPossibleVariableValues(
+                  node.value as Identifier,
+                  scopeManager,
+                );
 
           if (
             values.length === 0 ||
             values.some((val) => val.type !== 'Literal') ||
-            isVariableFromParameter(node.value, scopeManager)
+            isVariableFromParameter(node.value as Identifier, scopeManager)
```
bmish: Avoid casting.

lib/rules/no-unused-message-ids.ts
```diff
@@ -127,18 +135,23 @@ const rule = {
...
           ) {
             // When a dynamic messageId is used and we can't detect its value, disable the rule to avoid false positives.
             hasSeenUnknownMessageId = true;
           }
 
-          values.forEach((val) => messageIdsUsed.add(val.value));
+          values.forEach(
+            (val) => 'value' in val && messageIdsUsed.add(val.value as string),
+          );
```
bmish: Avoid casting.

lib/rules/no-useless-token-range.ts
```diff
@@ -51,33 +58,35 @@ const rule = {
         arg.properties.length >= 2 ||
         (arg.properties.length === 1 &&
           (getKeyName(arg.properties[0]) !== 'includeComments' ||
-            arg.properties[0].value.type !== 'Literal'))
+            (arg.properties[0] as Property).value.type !== 'Literal'))
```
bmish: type check instead of casting

lib/rules/no-useless-token-range.ts
```diff
@@ -51,33 +58,35 @@ const rule = {
...
       );
     }
 
+    function isMemberExpression(node: Node): node is MemberExpression {
```
bmish: I believe we can avoid simple type-checking functions like this. Inlining the type check should work fine for narrowing the type.

lib/rules/no-useless-token-range.ts
```diff
@@ -139,20 +150,22 @@ const rule = {
...
+                : identifier.parent.parent.parent;
             const replacementText =
               sourceCode.text.slice(
-                fullRangeAccess.range[0],
-                identifier.parent.parent.range[0],
+                fullRangeAccess.range![0],
+                identifier.parent.parent.range![0],
+              ) +
+              sourceCode.getText(
+                (identifier.parent.parent as CallExpression).arguments[0],
```
bmish: Type check instead of casting.

lib/rules/no-useless-token-range.ts
```diff
@@ -161,7 +174,9 @@ const rule = {
               fix(fixer) {
                 return fixer.replaceText(
                   identifier.parent.parent,
-                  sourceCode.getText(identifier.parent.parent.arguments[0]),
+                  sourceCode.getText(
+                    (identifier.parent.parent as CallExpression).arguments[0],
```
bmish: Type check instead of casting.

lib/rules/report-message-format.ts
```diff
@@ -41,23 +40,25 @@ const rule = {
...
       if (
         (message.type === 'Literal' &&
           typeof message.value === 'string' &&
           !pattern.test(message.value)) ||
         (message.type === 'TemplateLiteral' &&
           message.quasis.length === 1 &&
-          !pattern.test(message.quasis[0].value.cooked)) ||
-        (staticValue && !pattern.test(staticValue.value))
+          !pattern.test(message.quasis[0].value.cooked ?? '')) ||
+        (staticValue && !pattern.test(staticValue.value as string))
```
bmish: Type check instead of casting.

lib/rules/require-meta-default-options.ts
```diff
@@ -68,7 +68,7 @@ const rule = {
 
     if (!metaDefaultOptions) {
       context.report({
-        node: metaNode,
+        node: metaNode!,
```
bmish: Type check instead of casting.

lib/rules/require-meta-docs-url.ts
```diff
@@ -102,7 +102,7 @@ const rule = {
           return;
         }
 
-        if (isExpectedUrl(staticValue && staticValue.value)) {
+        if (isExpectedUrl(staticValue && (staticValue.value as string))) {
```
bmish: Type-checking instead of casting.

lib/rules/require-meta-fixable.ts
```diff
@@ -99,7 +98,9 @@ const rule = {
           }
 
           if (
-            !['code', 'whitespace', null, undefined].includes(staticValue.value)
+            !['code', 'whitespace', null, undefined].includes(
+              staticValue.value as string,
```
bmish: Type-checking instead of casting.

lib/rules/require-meta-fixable.ts
```diff
@@ -111,7 +112,7 @@ const rule = {
 
           if (
             usesFixFunctions &&
-            !['code', 'whitespace'].includes(staticValue.value)
+            !['code', 'whitespace'].includes(staticValue.value as string)
```
bmish: Type-checking instead of casting.

lib/rules/require-meta-fixable.ts
```diff
@@ -121,7 +122,7 @@ const rule = {
           } else if (
             catchNoFixerButFixableProperty &&
             !usesFixFunctions &&
-            ['code', 'whitespace'].includes(staticValue.value)
+            ['code', 'whitespace'].includes(staticValue.value as string)
```
bmish: Type-checking instead of casting.

lib/rules/require-meta-type.ts
```diff
@@ -68,7 +62,7 @@ const rule = {
           return;
         }
 
-        if (!VALID_TYPES.has(staticValue.value)) {
+        if (!VALID_TYPES.has(staticValue.value as string)) {
```
bmish: Type-checking instead of casting.

lib/utils.ts
```diff
@@ -116,14 +161,19 @@ function isTypeScriptRuleHelper(node) {
...
+  scopeManager: Scope.ScopeManager,
+): PartialRuleInfo {
+  const possibleNodes: Node[] = [];
 
   for (const statement of ast.body) {
     switch (statement.type) {
       // export default rule;
       case 'ExportDefaultDeclaration': {
-        possibleNodes.push(statement.declaration);
+        possibleNodes.push(statement.declaration as Identifier);
```
bmish: Type-check instead of casting.

lib/utils.ts
```diff
@@ -141,23 +191,23 @@ function getRuleExportsESM(ast, scopeManager) {
           const nodes =
             statement.declaration.type === 'VariableDeclaration'
               ? statement.declaration.declarations.map(
-                  (declarator) => declarator.init,
+                  (declarator) => declarator.init!,
```
bmish: Instead of forcing this to exist, can we filter out null/undefined afterward with `.filter(val=>!!val)`?

lib/utils.ts
```diff
@@ -196,34 +246,37 @@ function getRuleExportsESM(ast, scopeManager) {
...
   let exportsIsFunction = false;
   return ast.body
     .filter((statement) => statement.type === 'ExpressionStatement')
     .map((statement) => statement.expression)
     .filter((expression) => expression.type === 'AssignmentExpression')
     .filter((expression) => expression.left.type === 'MemberExpression')
-
-    .reduce((currentExports, node) => {
+    .reduce<PartialRuleInfo>((currentExports, node) => {
+      const leftExpression = node.left as MemberExpression;
```
bmish: Type-check instead of casting.

lib/utils.ts
```diff
@@ -319,15 +381,15 @@ function findVariableValue(node, scopeManager) {
  * If a ternary conditional expression is involved, retrieve the elements that may exist on both sides of it.
  * Ex: [a, b, c] will return [a, b, c]
  * Ex: foo ? [a, b, c] : [d, e, f] will return [a, b, c, d, e, f]
- * @param {Node} node
- * @returns {Node[]} the list of elements
+ * @param node
```
bmish:
```suggestion
```

lib/utils.ts
```diff
@@ -399,19 +468,25 @@ export function getContextIdentifiers(scopeManager, ast) {
   return new Set(
     scopeManager
       .getDeclaredVariables(ruleInfo.create)
-      .find((variable) => variable.name === ruleInfo.create.params[0].name)
+      .find(
+        (variable) =>
+          variable.name === (ruleInfo.create.params[0] as Identifier).name,
```
bmish: Type-check instead of casting.

lib/utils.ts
```diff
@@ -420,7 +495,7 @@ export function getKeyName(property, scope) {
       // Variable key: { [myVariable]: 'hello world' }
       if (scope) {
         const staticValue = getStaticValue(property.key, scope);
-        return staticValue ? staticValue.value : null;
+        return staticValue ? (staticValue.value as string) : null;
```
bmish: Type-check instead of casting.

lib/utils.ts
```diff
@@ -861,47 +983,61 @@ export function getMessagesNode(ruleInfo, scopeManager) {
...
+  scopeManager: Scope.ScopeManager,
+): (Property | SpreadElement)[] | undefined {
   const messagesNode = getMessagesNode(ruleInfo, scopeManager);
 
   return messagesNode && messagesNode.type === 'ObjectExpression'
     ? evaluateObjectProperties(messagesNode, scopeManager)
     : undefined;
 }
 
+const isProperty = (node: Node): node is Property => node.type === 'Property';
```
bmish: I don't think having this helper is necessary?

package.json
```diff
@@ -3,28 +3,30 @@
...
     "lint:docs": "markdownlint \"**/*.md\"",
-    "lint:eslint-docs": "npm-run-all \"update:eslint-docs -- --check\"",
+    "lint:eslint-docs": "npm-run-all -s build \"update:eslint-docs -- --check\"",
     "lint:js": "eslint --cache --ignore-pattern \"**/*.md\" .",
     "lint:js-docs": "eslint --no-inline-config \"**/*.md\"",
     "lint:package-json": "npmPkgJsonLint .",
     "release": "release-it",
     "test": "vitest run --coverage",
-    "test:remote": "eslint-remote-tester -c ./eslint-remote-tester.config.ts",
+    "test:remote": "eslint-remote-tester",
```
bmish, line 23: Does it detect the TS config file automatically now?

tests/lib/rules/require-meta-type.ts
```diff
@@ -66,16 +66,6 @@ ruleTester.run('require-meta-type', rule, {
...
-    {
-      code: `
-        const create = {};
-        module.exports = {
-          meta: {},
-          create,
-        };
-      `,
-      errors: [{ messageId: 'missing' }],
-    },
```
bmish, line 78: What was this removed for?

types/estree.d.ts
bmish, line 1: What's the purpose of this file?

.gitignore
michaelfaith: I updated this, and added slashes to the other folders that didn't have them (e.g. `.idea/`, `coverage/` etc)

lib/rules/fixer-return.ts
michaelfaith: Updated.

lib/rules/meta-property-ordering.ts
michaelfaith: But that's how Map types already work, without having to define that in the generic.  Without changing the generic, here's what getting from `orderMap` shows as the type.  It already includes `undefined` 

image

bmish: Oh nice! Did not realize that. Thanks.

lib/rules/no-identical-tests.ts
michaelfaith: Updated

lib/rules/no-meta-schema-default.ts
michaelfaith: That sufficiently narrows the true condition, but not the false condition.  I'd still need something more to use `key.value`
image

lib/rules/no-missing-placeholders.ts
michaelfaith: the filter doesn't do enough to narrow the type here, unfortunately.  Since the filter ensures `message` has a value, I thought it'd be ok assert here.  There shouldn't ever be a case where `message` is falsey, but I can add the condition.

michaelfaith: Updated

lib/rules/no-only-tests.ts
michaelfaith: Updated

lib/rules/no-property-in-node.ts
michaelfaith: Updated

lib/rules/no-unused-message-ids.ts
michaelfaith: That's enough to narrow `val` to Literal, but it doesn't narrow `value` to string, unfortunately. 
image

This worked, though it's a bit verbose...
```ts
values
    .filter(value => value.type === 'Literal')
    .map(value => value.value)
    .filter(value => typeof value === 'string')
    .forEach(value => messageIdsUsed.add(value));
```

michaelfaith: The type narrowing in `filter` still isn't perfect 😓

michaelfaith:
I've started going through the feedback, but wanted to double check something with you. I tried to avoid changing the logic much from the original, to avoid inadvertently introducing bugs, which is why there's casting in a number of places (what I was referring to in the description).  In order to drop the casting on a lot of these, I'll need to add additional conditions and the like to narrow the types enough.  Just wanted to confirm that's the direction you want me to go?

types/estree.d.ts
michaelfaith: Extending / augmenting the estree node types to:

1. Add `parent` to all nodes, like ESLint does
2. Add the two TypeScript node types that we use in the code.  Since there's only use of those two, it didn't seem worth adding a dependency on TypeScript ESLint in this change, since it would be a bigger change.

tests/lib/rules/require-meta-type.ts
michaelfaith: This was in the `valid` array, but had an `errors` property, which isn't correct.  Should I add it to the invalid array?

lib/rules/no-missing-placeholders.ts
bmish: Yeah generally would prefer to use an early return even if not technically necessary. This helps avoid issues that casting can hide.

However, it looks like an inline type guard could also help here:

```ts
arr
  .filter((val): val is string => val !== undefined)
  .forEach(val => {
    // val is string
    console.log(val.toUpperCase());
  });
```

lib/rules/no-unused-message-ids.ts
bmish: Nice!

tests/lib/rules/require-meta-type.ts
bmish: I'm not sure if it's valid or invalid, but whichever one it is, it would be good to put it there.

bmish:
> I've started going through the feedback, but wanted to double check something with you. I tried to avoid changing the logic much from the original, to avoid inadvertently introducing bugs, which is why there's casting in a number of places (what I was referring to in the description). In order to drop the casting on a lot of these, I'll need to add additional conditions and the like to narrow the types enough. Just wanted to confirm that's the direction you want me to go?

We have pretty good tests and high test coverage, so yes I would add early returns and do other minor refactoring needed to avoid casting when practical. But this is a big PR, so if any of these improvements get too big or complicated, it would be fine to save for another time. My goal is to avoid casting as much as possible and only use it when it's a pain to avoid it.

types/estree.d.ts
bmish: Can you add this as a comment to the file explaining what these types are for and where they come from?

tests/lib/rules/test-case-shorthand-strings.ts
```diff
@@ -12,10 +12,10 @@ import { RuleTester } from 'eslint';
 
 /**
  * Returns the code for some valid test cases
- * @param {string[]} cases The code representation of valid test cases
+ * @param cases The code representation of valid test cases
  * @returns {string} Code representing the test cases
```
bmish:
```suggestion
 * @returns Code representing the test cases
```

lib/rules/no-unused-message-ids.ts
michaelfaith: Updated

lib/rules/no-unused-message-ids.ts
michaelfaith: Updated

lib/rules/no-unused-message-ids.ts
michaelfaith: Updated

lib/rules/no-unused-message-ids.ts
michaelfaith: Updated to 
```ts
values
            .filter((val) => val.type === 'Literal')
            .map((val) => val.value)
            .filter((val) => typeof val === 'string')
            .forEach((val) => messageIdsUsed.add(val));
```

tests/lib/rules/require-meta-type.ts
michaelfaith: Cool.  Based on the rule, I _think_ code should be invalid.  So, I'll add it to the `invalid` array.

michaelfaith: Oh, actually, nevermind.  Its `create` doesn't have a `context`, so it's not detecting it as a rule.  I.e. `valid`.  I'll add it back to valid and remove the `errors`.

michaelfaith: Done

types/estree.d.ts
michaelfaith: Added this comment
```ts
/**
 * This file augments the `estree` types to include a couple of types that are not built-in to `estree` that we're using.
 * This is necessary because the `estree` types are used by ESLint, and ESLint does not natively support
 * TypeScript types.  Since we're only using a couple of them, we can just add them here, rather than
 * installing typescript estree types.
 *
 * This also adds support for the AST mutation that ESLint does to add parent nodes.
 */
```

package.json
michaelfaith: It does!  😊

tests/lib/rules/test-case-shorthand-strings.ts
michaelfaith: Updated

lib/rules/no-meta-schema-default.ts
michaelfaith: This was actually an easy fix.  We'd already done the hard work, right above this to create `staticKey`, and then didn't use it in the `switch`.  So, changed the switch to use `staticKey.value`.

lib/rules/no-useless-token-range.ts
michaelfaith: This is another example of where we're already doing the check in the filter, above, it's just not narrowing that prop downstream, so it feels redundant.  But I'll go ahead and change it. 

image

lib/rules/no-useless-token-range.ts
michaelfaith: Updated

lib/rules/no-useless-token-range.ts
michaelfaith: Updated.  This was just for convenience, since there were so many.

lib/rules/no-useless-token-range.ts
michaelfaith: Updated

lib/rules/no-useless-token-range.ts
michaelfaith: Updated

lib/rules/report-message-format.ts
michaelfaith: Updated

lib/rules/require-meta-default-options.ts
michaelfaith: Updated

lib/rules/require-meta-docs-url.ts
michaelfaith: Updated

lib/rules/require-meta-fixable.ts
michaelfaith: Updated

lib/rules/require-meta-fixable.ts
michaelfaith: Updated

lib/rules/require-meta-fixable.ts
michaelfaith: Updated

lib/rules/require-meta-type.ts
michaelfaith: Updated

lib/utils.ts
michaelfaith: Updated

lib/index.ts
```diff
@@ -41,20 +39,56 @@ import requireMetaType from './rules/require-meta-type.js';
 import testCasePropertyOrdering from './rules/test-case-property-ordering.js';
 import testCaseShorthandStrings from './rules/test-case-shorthand-strings.js';
 
+const require = createRequire(import.meta.url);
+
+const packageMetadata = require("../package.json") as {
```
bmish, line 44: I had issues with this which I fixed in https://github.com/qunitjs/eslint-plugin-qunit/pull/584

michaelfaith: That doesn't really apply in this case, since this file's relative position to the package.json is the same in the dist folder as it is in the lib folder.

lib/utils.ts
michaelfaith: Updated

lib/utils.ts
michaelfaith: Updated

lib/utils.ts
michaelfaith: Updated

lib/utils.ts
michaelfaith: Updated

lib/utils.ts
michaelfaith: Updated

lib/utils.ts
michaelfaith: Updated

michaelfaith:
I believe I've addressed all of the feedback so far.  Let me know if there are any other updates you think should be made. Appreciate the thorough review.

.github/workflows/main.yml
```diff
@@ -45,3 +45,13 @@ jobs:
...
       - run: npm run test:remote
+
+  typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+      - run: npm install
```
bmish, line 56: Technically, I believe we should be using `npm ci` in these workflows. Fine to consider it as a follow-up.

lib/index.ts
bmish: Nice! Looks like I should be able to move index.js to fix that in my project:
- https://github.com/qunitjs/eslint-plugin-qunit/pull/619

lib/rules/require-meta-docs-recommended.ts
```diff
@@ -1,31 +1,31 @@
...
+) {
   if (objectNode.properties.length === 0) {
     return fixer.replaceText(
       objectNode,
       `{ recommended: ${recommendedValue} }`,
     );
   }
   return fixer.insertTextAfter(
-    objectNode.properties.at(-1),
+    objectNode.properties.at(-1)!,
```
bmish: One way to work around this is to do:

```ts
const lastProperty = objectNode.properties.at(-1);
if (!lastProperty) {
  return ...
}
return fixer.insertTextAfter(lastProperty, ...);

File: lib/utils.ts [hunk 37]
```diff
@@ -694,27 +800,33 @@ export function getSourceCodeIdentifiers(scopeManager, ast) {
...
+  node: ObjectExpression,
+  propertyText: string,
+  sourceCode: SourceCode,
+): Rule.Fix {
   if (node.properties.length === 0) {
     return fixer.replaceText(node, `{\n${propertyText}\n}`);
   }
   return fixer.insertTextAfter(
-    sourceCode.getLastToken(node.properties.at(-1)),
+    sourceCode.getLastToken(node.properties.at(-1)!)!,
```
bmish: One way to work around this is to do:

```ts
const lastProperty = objectNode.properties.at(-1);
if (!lastProperty) {
  return ...
}
return fixer.insertTextAfter(lastProperty, ...);

File: package.json [hunk 38]
```diff
@@ -75,6 +81,7 @@
     "npm-run-all2": "^7.0.1",
     "prettier": "^3.4.1",
     "release-it": "^17.2.0",
+    "tsup": "^8.5.0",
     "typescript": "^5.8.3",
```
bmish, line 85: In this or a later PR, we can upgrade this:
```suggestion
    "typescript": "^5.9.2",
```

tests/lib/utils.ts
```diff
@@ -1606,12 +1822,28 @@ describe('utils', () => {
...
+                  (
+                    (
+                      (ast.body[0] as ExpressionStatement)
+                        .expression as AssignmentExpression
+                    ).right as ObjectExpression
+                  ).properties[0] as Property
+                ).value as ObjectExpression
+              ).properties[0] as Property
+            ).value as ObjectExpression
+          ).properties as Property[];
```
bmish, line 1846: Just wondering what's going on here?

bmish (changes requested):
It's looking good. Just a few more fixes.

Would you also mind just running some manual sanity checks on the compiled output? E.g. try loading the plugin from `dist/`. 

And make sure `dist/` is ignored by tools like the following PRs (issues I discovered during the build/publishing process):
- https://github.com/qunitjs/eslint-plugin-qunit/pull/585
- https://github.com/qunitjs/eslint-plugin-qunit/pull/602

package.json
michaelfaith: The version of `@typescript-eslint` that's installed doesn't include 5.9 in the peer deps yet.  I'd suggest having a separate change that updates both sets of packages.  I don't mind doing that one too.

lib/utils.ts
michaelfaith: Updated

tests/lib/utils.ts
michaelfaith: That's constructing the AST types that the tests all assume, based on the text of the string.  It was extremely tedious, using `AST Explorer` and the test strings to figure it out 😅.  There's not a great way to infer the types of what a string's AST will ultimately be parsed into, yet the tests all assume the structure of the string of the test case.  Definite opportunity for a large refactor of how these tests operate.

lib/rules/require-meta-docs-recommended.ts
michaelfaith: Updated

.github/workflows/main.yml
michaelfaith: Yeah, I just matched the other steps.   Could be a follow-up that changes them all to ci?

bmish (approved):
Awesome work, thank you!

.github/workflows/main.yml
michaelfaith: I was thinking a bit more on this later.  In order to use `npm ci`, it'll require than just updating the calls in the workflow.  The repo would need to start maintaining a lock file.  It currently has lock files explicitly disabled, and `npm ci` requires that.  I personally don't think having a lock file is a bad thing.  I was kind of surprised when I first realized there wasn't one.  But it seemed to be a conscious choice someone made at some point.  

The other consideration is that using `npm ci` in the builds will result in longer build times, if that's important to you.

bmish: Oh good catch. I now remember that's why we don't use `npm ci`. Let's avoid adding it at this time. For us, it's fine to be more aggressive with dependency updates and omit a lockfile.

PR #535, opened 2025-08-04
build: update typescript to 5.9

michaelfaith:
This change updates `typescript` to 5.9

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "prettier": "^3.4.1",
     "release-it": "^17.2.0",
     "tsup": "^8.5.0",
-    "typescript": "~5.9.2",
+    "typescript": "^5.9.2",
     "vitest": "^3.2.4"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -74,15 +74,15 @@
...
+    "jiti": "^2.5.1",
     "lodash": "^4.17.21",
     "markdownlint-cli": "^0.43.0",
     "npm-package-json-lint": "^8.0.0",
     "npm-run-all2": "^7.0.1",
     "prettier": "^3.4.1",
     "release-it": "^17.2.0",
     "tsup": "^8.5.0",
-    "typescript": "^5.8.3",
+    "typescript": "~5.9.2",
```
bmish: Let's stick with caret. We can be more aggressive with dependency updates in this repo.

bmish: I've updated the PR. If we want to change our minds about caret, we can consider it separately.

bmish (approved):
Thanks!

PR #540, opened 2025-08-04
build: migrate prettier config to typescript and remove eslint-plugin-prettier

michaelfaith:
This change moves from using `eslint-prettier-plugin` to using `eslint-config-prettier` and prettier directly for format checking.

A result of us never running prettier on the root files, is that there were some adjustments needed.  I ran prettier on the repo and have added those changes too.

Lastly, I moved the config to `typescript`.

### DIFF
--- a/tests/lib/utils.ts
+++ b/tests/lib/utils.ts
@@ -1653,8 +1653,8 @@ describe('utils', () => {
     });
   });
 
-  describe('evaluateObjectProperties', function () {
-    it('behaves correctly with simple object expression', function () {
+  describe('evaluateObjectProperties', () => {
+    it('behaves correctly with simple object expression', () => {
       const getObjectExpression = (ast: Program): ObjectExpression =>
         (ast.body[0] as VariableDeclaration).declarations[0]
           .init as ObjectExpression;
@@ -1670,7 +1670,7 @@ describe('utils', () => {
       assert.deepEqual(result, getObjectExpression(ast).properties);
     });
 
-    it('behaves correctly with spreads of objects', function () {
+    it('behaves correctly with spreads of objects', () => {
       const getObjectExpression = (
         ast: Program,
         bodyElement: number,
@@ -1703,7 +1703,7 @@ describe('utils', () => {
       ]);
     });
 
-    it('behaves correctly with non-variable spreads', function () {
+    it('behaves correctly with non-variable spreads', () => {
       const getObjectExpression = (ast: Program): ObjectExpression =>
         (ast.body[1] as VariableDeclaration).declarations[0]
           .init as ObjectExpression;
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -68,8 +70,7 @@
     "eslint-doc-generator": "^2.2.2",
     "eslint-plugin-markdown": "^5.1.0",
     "eslint-plugin-n": "^17.21.0",
-    "eslint-plugin-prettier": "^5.5.3",
-    "eslint-plugin-unicorn": "^56.0.1",
+    "eslint-plugin-unicorn": "^60.0.0",
```
bmish: If there are a lot of fixes from the unicorn upgrade, it would be preferred to do it separately.

michaelfaith: Reverted the unicorn plugin update

michaelfaith:
This is failing because the `setup-node` action is still pulling 22.17.1, instead of 22.18.0, which doesn't have the type stripping.  

image

I'm going to add `check-latest: true` to that steps action, to force it to check for the latest node version

michaelfaith:
That worked 

image

But there's still an issue with one of the lint commands.  Looking into it

eslint.config.ts
```diff
@@ -25,10 +27,10 @@ export default tseslint.config([
   ...compat.extends(
     'not-an-aardvark/node',
     'plugin:@eslint-community/eslint-comments/recommended',
-    'plugin:prettier/recommended',
     'plugin:unicorn/recommended',
   ),
   pluginN.configs['flat/recommended'],
+  prettier,
```
bmish, line 33: I think we technically want to include this last to ensure it overrides any incompatible config?

> eslint.config.js (flat config): Import eslint-config-prettier, and put it in the configuration array – after other configs that you want to override.

package.json
```diff
@@ -12,10 +12,12 @@
   "license": "MIT",
   "scripts": {
     "build": "tsup",
+    "format": "prettier --write .",
+    "format:check": "prettier --check .",
     "lint": "npm-run-all --continue-on-error --aggregate-output --parallel lint:*",
     "lint:docs": "markdownlint \"**/*.md\"",
     "lint:eslint-docs": "npm-run-all -s build \"update:eslint-docs -- --check\"",
-    "lint:js": "eslint --cache --ignore-pattern \"**/*.md\" .",
```
bmish, line 18: Isn't the path needed?

eslint.config.ts
michaelfaith: I will generally put the prettier config after all of the reusable configs, but before any configs that I've created with overrides / additional rules.  The idea being that anything I add myself is an intentional choice and should take precedence, which I think aligns with the spirit of that documentation.  With that said, it's currently before the TS Configs, so I should at least move it to be after those (I authored this PR before adding the TS configs).  Would you prefer I move it all the way to the end or just after the ts configs?

package.json
michaelfaith: No, it's optional.  It defaults to current working directory.  I can add it back, if you prefer? 

https://eslint.org/docs/latest/use/command-line-interface
Image

bmish: Fine to use default then.

eslint.config.ts
bmish: Your reasoning makes sense. I'll leave it up to you.

bmish (approved):
Thanks!

PR #541, opened 2025-08-05
refactor: update all relative export paths to use ts extension

michaelfaith:
Coming from https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/540.  This change shifts all relative imports to use `.ts` extensions, and enables `rewriteRelativeImportExtensions` in the tsconfig.  Unfotunately `tsup` has a bug when using this setting (https://github.com/egoist/tsup/issues/1271), so I switched bundlers to `tsdown` the Vite community's new `tsup` alternative (which does support this).

I also updated the lint CI step to force using the latest LTS (22.18.0) which comes with native type stripping.  I also added the eslint flag that allows eslint to use native type stripping.

### DIFF
--- a/tsdown.config.ts
+++ b/tsdown.config.ts
@@ -2,9 +2,7 @@ import { defineConfig } from 'tsdown';
 
 export default defineConfig({
   unbundle: true,
-  clean: true,
   dts: true,
   entry: ['lib/**/*.ts'],
-  format: ['esm'],
   outDir: 'dist',
 });


bmish:
Are `.ts` imports desirable for us regardless of these tooling choices?

michaelfaith:
> Are `.ts` imports desirable for us regardless of these tooling choices? Any other pros and cons to switching to them?

The two main motivators i see are:

1. Allows use of node's native type stripping
2. More clear to newer devs, since the extension matches the imported source file's extension.

package.json
```diff
@@ -11,14 +11,14 @@
...
+    "build": "tsdown",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "lint": "npm-run-all --continue-on-error --aggregate-output --parallel lint:*",
     "lint:docs": "markdownlint \"**/*.md\"",
     "lint:eslint-docs": "npm-run-all -s build \"update:eslint-docs -- --check\"",
-    "lint:js": "eslint --cache --ignore-pattern \"**/*.md\"",
-    "lint:js-docs": "eslint --no-inline-config \"**/*.md\"",
+    "lint:js": "eslint --cache --ignore-pattern \"**/*.md\" --flag unstable_native_nodejs_ts_config",
+    "lint:js-docs": "eslint --no-inline-config \"**/*.md\" --flag unstable_native_nodejs_ts_config",
```
bmish, line 21: Can we get rid of `jiti` if we're using this flag?

bmish:
Just want to verify the output would still be compatible with the minimum Node versions we support? (`20.19.0`, `22.13.1` which don't support type-stripping yet). May be worth adding these versions to the CI workflow for running tests.

package.json
michaelfaith: We wouldn't need it for the `eslint` run anymore. But it’s still being used by `eslint-remote-tester` to parse that config. And, unfortunately, they don't have an option to use node's built-in, like eslint does.

michaelfaith:
> Just want to verify the output would still be compatible with the minimum Node versions we support? (`20.19.0`, `22.13.1` which don't support type-stripping yet). May be worth adding these versions to the CI workflow for running tests.

Yes, the output tsdown generates still has .js extensions, and this wouldn't change the compatability of what it emits. I can prepare a before / after diff to share?

Unfortunately, adding those versions to the test matrix wouldn't really be an adequate test, since the tests are run on source rather than the build output. But i like that idea.

bmish (approved):
Thanks!

tsdown.config.ts
```diff
@@ -1,10 +1,8 @@
-import { defineConfig } from 'tsup';
+import { defineConfig } from 'tsdown';
 
 export default defineConfig({
-  bundle: false,
-  clean: true,
+  unbundle: true,
   dts: true,
   entry: ['lib/**/*.ts'],
-  format: ['esm'],
```
bmish, line 8: Are there any behavior changes from this tsup/tsdown conversion?

michaelfaith: No, it's meant to be a drop in replacement, just faster. If we were able to set `isolatedModules` to true in the tsconfig, it would be even faster. But the way the plugin is setup, we can't really do that, and still get rich type information. So, yeah, just swapping one for the other, won't change anything. I was able to remove some of those settings because tsdown will infer them from the package.json.

PR #542, opened 2025-08-05
build: set up typescript-eslint

### LINKED ISSUE #543: Add typescript-eslint to repo lint config
Following the migration of the plugin to typescript, the eslint config needs to be updated to use the `typescript-eslint` plugin.

michaelfaith:
This change adds typescript-eslint to our config and fixes a miss in the original TypeScript migration where we weren't linting the newly renamed `ts` files.

This also addresses violations found from both sets of config changes.

Closes #543

### DIFF
--- a/docs/rules/no-property-in-node.md
+++ b/docs/rules/no-property-in-node.md
@@ -13,7 +13,8 @@ Instead, checking a node's `type` property is generally considered preferable.
 
 Examples of **incorrect** code for this rule:
 
-```ts
+```js
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     /* ... */
@@ -27,12 +28,13 @@ module.exports = {
       },
     };
   },
-} satisfies Rule.RuleModule;
+};
 ```
 
 Examples of **correct** code for this rule:
 
-```ts
+```js
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     /* ... */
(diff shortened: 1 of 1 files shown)


docs/rules/no-property-in-node.md
```diff
@@ -25,20 +22,17 @@ module.exports = {
...
       'ClassDeclaration, FunctionDeclaration'(node) {
         if ('superClass' in node) {
-          console.log('This is a class declaration:', node);
+          // This is a class declaration
         }
       },
     };
   },
-};
+} satisfies Rule.RuleModule;
```
bmish: It seems like we're mixing commonjs syntax and TypeScript in this example, which is a bit odd. Can we choose one:
- commonjs with the jsdoc types
- or ESM TypeScript and include any imports for the types

michaelfaith: Definitely agree with you there.  The docs are a bit all over the place, with different people writing them at different points in time.  It would benefit from someone going through all of them, with a focus on making them consistent (which I'm also happy to do).    I don't think that's really in the scope of this change, though, which is just focused on enabling the eslint plugin.  I had to touch this file, to address a violation. Would you prefer I change it to js here, and remove the `satisfies`?

bmish: Leaving it as cjs would be best then.

michaelfaith: Updated.  I changed the examples to `js` and put the type annotation back in.

bmish (approved):
Thanks!

PR #546, opened 2025-08-10
ci: add e2e tests to CI

michaelfaith:
This change adds a new end-to-end test capability.  There are two new jobs in the main CI workflow.  One job to build the package (which also good to have outside of the need to run e2e tests), executed with the latest LTS node.  The build artifact is uploaded for the downstream `e2e` job to use.  The `e2e` job, downloads the artifact, run `npm install` and then `npm run e2e`, which executes a custom node script to loop through subdirectories of `/e2e`, which each contain a project fixture with different setups.  Both projects have the same simple plugin source code, and each has this plugin (referenced from the root of the repo (aka the result of the build)) and `eslint` installed as `devDependencies`.  Each also has a `lint` package.json script that runs `eslint` on the simple plugin.

## `/e2e/all`

This fixture has a regular js config and runs our `all` config on the project.

## `/e2e/all-typed-config`

This fixture has `typescript` installed and uses a ts-based config and runs our `all` config on the project.  It also executes `tsc` as part of the `lint` command, to ensure everything's typed property.


e2e/all/.npmrc
bmish, line 1: I think we need a clear `fixtures` directory to contain these inside. `e2e/fixtures/...`.

eslint.config.ts
```diff
@@ -18,7 +18,7 @@ const compat = new FlatCompat({
 export default tseslint.config([
   // Global ignores
   {
-    ignores: ['node_modules', 'coverage', 'dist', 'tests/lib/fixtures'],
+    ignores: ['node_modules', 'coverage', 'dist', 'tests/lib/fixtures', 'e2e'],
```
bmish: Should we only ignore the fixtures?

.npmpackagejsonlintignore
```diff
@@ -0,0 +1 @@
+e2e/
```
bmish: newline at end of file:

```suggestion
e2e/

```

e2e/runAllTests.js
bmish, line 1: A summary comment at the top of the file explaining how this e2e testing works could be helpful.

e2e/runAllTests.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+    try {
+      execSync(TEST_COMMAND, {
+        cwd: testDir,
+        stdio: ['ignore', 'ignore', 'pipe'],
+      });
+      console.log(`✅ Test passed\n`);
+    } catch (error) {
+      console.log(`❌ Test failed`);
+      console.error(`${error}\n`);
```
bmish: Will this print the actual lint violations to help with debugging? I don't see the actual errors in the screenshot. This means someone would have to rerun locally which is not ideal.

.github/workflows/main.yml
```diff
@@ -67,3 +67,46 @@ jobs:
...
+      matrix:
+        node-version:
+          - '24.0.0' # minimum supported v24
+          - 24
+          - '22.13.1' # minimum supported v22
+          - 22
+          - '20.19.0' #minimum supported v20
+          - 20
+        os:
+          - ubuntu-latest
```
bmish, line 100: Looks like we may be missing tests on Windows. Should we test that here?

.github/workflows/main.yml
```diff
@@ -67,3 +67,46 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
```
bmish, line 106: Can we use `cache: 'npm'` on most of these setup-node actions in the file?

bmish (commented):
Neat.

I also ran this through AI if you'd like to have a look at what it thought: https://chatgpt.com/share/68993709-d248-800c-a59f-c765c9b7b043

e2e/all/.npmrc
michaelfaith: would the goal be to allow us to not ignore the run script in the esint config?  I was thinking of possibly adding a `fixtures` do so that there could be an `e2e/package.json` that's set up with workspaces, so that there only has to be one `npm install` call to pull in dependencies for all fixtures?  What do you think?

e2e/runAllTests.js
michaelfaith: That's a good point.  It's only outputting which script command failed. 
image

But not what caused the command to fail.  I'll improve that

.github/workflows/main.yml
michaelfaith: Sure, I can add that.  Do you want it added to the unit test matrix too?

.github/workflows/main.yml
michaelfaith: Updated

michaelfaith: Ah, looks like a lock file is necessary to use the `cache` option.  Pulling that back out, unless you want me to add a lock file?

.github/workflows/main.yml
michaelfaith: Added to the e2e job.  If you want me to add it to the test job, I can do that too.

.npmpackagejsonlintignore
michaelfaith: Updated

e2e/runAllTests.js
michaelfaith: What do you think about this?  

image

e2e/runAllTests.js
michaelfaith: Added

e2e/all/.npmrc
bmish: I'd like to see fixtures moved to a `fixtures/` directory to clearly designate what's live code and what's not. It's also useful for easily ignoring fixtures or applying other fixture-only config.

In terms of workspaces, I'm open to it if you think there's a worthwhile benefit i.e. making our setup simpler and cleaner. At the same time, I'd prefer to avoid overcomplicating our setup and increasing the maintenance burden.

.github/workflows/main.yml
bmish: Ah yes. Let's avoid it then.

e2e/runAllTests.js
bmish: Seems fine.

.github/workflows/main.yml
bmish: Generally would be good to test on both. Hopefully it doesn't hurt perf too much. Also could be done separately.

e2e/runAllTests.js
```diff
@@ -0,0 +1,63 @@
+import { execSync } from 'node:child_process';
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import process from 'node:process';
+
+/**
+ * Run All Tests: This script executes the lint command on all subfolders under `fixtures`, in order
+ * to validate the correctness of our plugin.  Each fixture installs the *built* package.  So this should
+ * only be run after a build has been done.
```
bmish: Out of curiosity, how is the performance impact of this on our builds?

michaelfaith: In total, it adds around 40s to a minute to the overall CI runtime. Mainly because the E2E steps have to wait for the build to complete.  Just looking at the build readout below, it took 38s for the build job, which ran in parallel to all of the other existing jobs.  At that point all the E2E jobs kicked off in parallel, and the longest running of those are the windows ones, which took about a minute each (the ubuntu ones took between 40-50s).

michaelfaith: RE: performance, though, there will be a roughly linear impact to the length of those steps, for each e2e test that's added in the future.  So, one thing to consider, if many more are added, would be to run some number of them concurrently.  I wouldn't necessarily go to that right now, because it adds some amount of complexity.  With just two tests, there's not much to gain with doing them concurrently.  But if there were 10 or more, it might start to make sense.

e2e/all/.npmrc
michaelfaith: Sounds good. I'll make that change.

michaelfaith: Updated

eslint.config.ts
michaelfaith: Updated

e2e/.npmrc
```diff
@@ -0,0 +1 @@
+package-lock = false
```
bmish: new line at end of file
```suggestion
package-lock = false

```

bmish:
> Btw, I'm thinking it might be a good idea to move the prettier config back to js, until the VSCode extension supports it. [prettier/prettier-vscode#3623](https://github.com/prettier/prettier-vscode/issues/3623) IDE-autoformatting isn't working anymore. What do you think?

Up to you on that. If it's going to be broken for a while, then could make sense to revert. If it's going to be fixed soonish, then could wait.

aladdin-add:
Is there a way to simplify this process? This change introduces 10+ jobs, and I wonder if it's really necessary - although grateful for Microsoft's generosity in offering it for free to open-source projects, I still prefer to minimize the resource consumption.

michaelfaith:
Idk, there's been a lot of changes recently to node that directly affect tooling (e.g. require(ESM), type stripping, etc).  And so, I think testing the various versions of node is important, since it can avoid blind spots.  RE: OSes, I know that we've had issues with Windows compatibility on some of our stuff in `eslint-plugin-package-json` in the past, and adding tests for windows helped avoid those from coming back.  To be honest, it feels like a low cost way of giving you confidence that the support that you're claiming is there, is actually there.  🤷‍♂️

PR #553, opened 2025-10-21
feat: add new rule `require-test-case-name`

### LINKED ISSUE #544: New Rule: `test-case-name-property`
## Proposal

I'd like to propose a new rule that require's the use of a `name` property in test cases that aren't simple strings.  I'm thinking the default behavior would be for it to require it, only if the test case object has `options` and / or `settings` defined.  If the object just has 'code`, then don't flag it.  And then perhaps there's also an option to require it always?

## Motivation

The default `Ruletester` output is usually sufficient when the test cases only consist of the code.  But if the test case includes options and / or settings, the code for that test case is likely shared with other test cases (that just have different options / settings applied).  In such cases, the output is still just the code.  So, if you're trying to track down a failure, and you only have the code from the failure to work with, it can be hard to find the culprit test case.

Here's an example of where we adopted this pattern in `eslint-plugin-package-json`

| Before | After |
| -------- | ------- |
| Image | Image  |

Thoughts?

michaelfaith:
This change adds a new rule to require `name` be present in test cases under certain conditions.  This rule aims to ensure test suites are producing logs in a form that make it easy to identify failing test, when they happen. For thoroughly tested rules, it's not uncommon to have the same `code` across multiple test cases, with only `options` or `settings` differing between them. Requiring these test cases to have a `name` helps ensure the test output is meaningful and distinct.

Closes #544

### DIFF
--- a/tests/lib/rules/require-test-case-name.ts
+++ b/tests/lib/rules/require-test-case-name.ts
@@ -55,6 +55,12 @@ ruleTester.run('test-case-name-property', rule, {
     getTestCases([
       '{ code: "foo", name: "my test", options: ["bar"], settings: { setting1: "bar" } }',
     ]),
+    getTestCases(['foo']),
+    getTestCases(['{ code: foo }']),
+    getTestCases(['foo()']),
+    getTestCases(['{ code: foo() }']),
+    // eslint-disable-next-line no-template-curly-in-string -- testing edge cases
+    getTestCases(['`foo;${bar}`']),
     getTestCases([], ['{ code: "foo", errors: ["some error"] }']),
     getTestCases(
       [],


lib/rules/require-test-case-name.ts
```diff
@@ -0,0 +1,141 @@
...
+        'This test case is missing the `name` property.  All test cases should have a name property.',
+      nameRequiredObjects:
+        'This test case is missing the `name` property.  Test cases defined as objects should have a name property.',
+      nameRequiredObjectsWithConfig:
+        'This test case is missing the `name` property.  Test cases defined as objects with additional configuration should have a name property.',
+    },
+  },
+
+  create(context) {
+    const { require: requireOption = 'objects-with-config' }: Options =
```
aladdin-add, line 65: 👍 it's needed as the package supports `eslint>=9.0.0`: https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/6a2f25d2bb564e455d80b919af59b9a734bb1505/package.json#L93

while `defaultOptions` was introduced in v9.15.0

lib/rules/require-test-case-name.ts
```diff
@@ -0,0 +1,141 @@
...
+      {
+        additionalProperties: false,
+        properties: {
+          require: {
+            description:
+              'When should the name property be required on a test case object.',
+            enum: ['always', 'objects', 'objects-with-config'],
+          },
+        },
+        type: 'object',
```
aladdin-add, line 50: Not quite sure if the object-style configuration is better than enumeration, what do you think?

```js
{ rules: {
  "require-test-case-name": ["error", "always"]
}}

vs.

{ rules: {
  "require-test-case-name": ["error", {require: "always"}]
}}
```

tests/lib/rules/require-test-case-name.ts
```diff
@@ -0,0 +1,446 @@
...
+
+const objectsWithConfigError = {
+  messageId: 'nameRequiredObjectsWithConfig',
+};
+
+const ruleTester = new RuleTester({
+  languageOptions: { sourceType: 'commonjs' },
+});
+ruleTester.run('test-case-name-property', rule, {
+  valid: [
```
aladdin-add, line 39: It would be even better if there were some tests covering non-string and non-object cases:

```
getTestCases(['foo']),
getTestCases(['{code: foo}']),
getTestCases(['foo()']),
getTestCases(['{code: foo()}']),
getTestCases(['`foo;${bar}`']),
```

aladdin-add (approved):
👍 Looks great!  just a few questions/minor suggestions.

Leaving it open 2~3 days for others to review.

tests/lib/rules/require-test-case-name.ts
michaelfaith: Makes sense. I'll add tests for these

lib/rules/require-test-case-name.ts
michaelfaith: I kind of lean towards using object in place of regular enum for extensibility.  If we ever wanted to add additional options to the rule, those wouldn't be breaking if the option type is already an object, and could be introduced in a minor release.  If it's a simple enum, then any additional options would have to be a breaking change.

tests/lib/rules/require-test-case-name.ts
michaelfaith: I added those additional tests.

lib/rules/require-test-case-name.ts
aladdin-add: agreed. I can see both sides: enum is a little easier to config. Both are fine. 👍

PR #567, opened 2025-11-04
feat: add new rule `no-matching-violation-suggest-message-ids`

### LINKED ISSUE #368: Rule: require suggestions have different `messageId` than their parent report
Code like this happens once in a while from contributors unfamiliar with ESLint rule suggestions:

```js
context.report({
  node,
  messageId: "someMessage",
  suggest: [
    {
      messageId: "someMessage",
      // ...
    },
  ],
});
```

At least in typescript-eslint we try to use different messages for suggestions. Suggestion messages are typically action-oriented, whereas parent report messages are statements of fact.

What do you think about a lint rule helping devs know to use a different message between the two?

StyleShit:
Closes #368

### DIFF
--- a/docs/rules/no-matching-violation-suggest-message-ids.md
+++ b/docs/rules/no-matching-violation-suggest-message-ids.md
@@ -65,5 +65,5 @@ module.exports = {
 - [ESLint rule docs: Providing Suggestions](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions)
 - [ESLint rule docs: Suggestion `messageId`s](https://eslint.org/docs/latest/extend/custom-rules#suggestion-messageids)
 - [no-missing-message-ids](./no-missing-message-ids.md) rule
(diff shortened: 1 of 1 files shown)


lib/rules/no-matching-violation-suggest-message-ids.ts
```diff
@@ -0,0 +1,128 @@
...
+      return {};
+    }
+
+    let contextIdentifiers: Set<Node>;
+
+    return {
+      Program(ast) {
+        contextIdentifiers = getContextIdentifiers(scopeManager, ast);
+      },
+
```
aladdin-add: you can just get the ast directly.
```suggestion
    const contextIdentifiers: Set<Node> = getContextIdentifiers(scopeManager, sourceCode.ast);

    return {

```

StyleShit: Cool, thanks!

StyleShit: Done

PR #574, opened 2026-01-03
ci: complete release-please update

### LINKED ISSUE #573: Tooling: release-please update not complete
Hey there 👋  [I recently migrated](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/pull/1478) a handful of repos I help @JoshuaKGoldberg maintain to use `release-please`, and this repo was one of my points of reference (along with `yargs` and a couple others), and I found that the update from v3 -> v4 that was done here a little while ago wasn't done all the way.  

As part of the major version bump, most of the action options that control the `release-please` behavior were moved from action props to a json config file: https://github.com/googleapis/release-please-action?tab=readme-ov-file#upgrading-from-v3-to-v4, however this repo still has them as action inputs.  If you look in the last run of the release-please workflow, you'll see the warnings about the invalid inputs: https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/19239812037/job/54999370466

Image

I believe it "just works" because you have the release-type set to `node` so it's just falling back to the node defaults, which are close to what you had in your settings.  But if you actually want the settings you've provided to be applied, they should be moved to the config file (you don't need `package-name` at all). 

https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/57cbe73a6616a1f64a9341e72f16568eac0fa228/.github/workflows/release-please.yml#L18-L31

I'm happy to help correct this, if you want. (cc @MichaelDeBoey)

michaelfaith:
This change completes the work started to update release-please from v3 to v4.  As part of that update, they moved most of the non-action related configuration options to a json config file, removing the action inputs.  I also updated the CI workflow so that adding a `ci` label to a release-please created PR will trigger CI for you, so you can have the ability to do a quick sanity check before merging the release.

Closes #573

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,8 +1,7 @@
 name: CI
 
 on:
-  pull_request:
-    types: [labeled, opened, reopened, synchronize]
+  pull_request: ~
   push:
     branches:
       - main


.github/workflows/release-please.yml
```diff
@@ -37,8 +21,7 @@ jobs:
           registry-url: https://registry.npmjs.org
         if: ${{ steps.release.outputs.release_created }}
       - run: |
-          npm install -g npm@latest  # npm 11.5.1 or later is required so update to latest to be sure
```
michaelfaith, line 40: Since node `lts/*` is being used, and lts is now v24, it already comes with a valid version of `npm` supporting trusted publishing.  This is no longer needed.

.github/workflows/release-please.yml
```diff
@@ -37,8 +21,7 @@ jobs:
...
           npm install --force
           npm run build
-          npm publish --provenance
```
michaelfaith, line 43: The Trusted Publishing config on the npm side handles provenance now.  No need to use this flag.

.github/workflows/main.yml
```diff
@@ -1,8 +1,11 @@
 name: CI
 
 on:
-  - push
-  - pull_request
+  pull_request:
+    types: [labeled, opened, reopened, synchronize]
```
aladdin-add: I'm not quite sure if it's necessary to trigger CI again when adding labels- is there any specific necessity for that?

michaelfaith: Since there's no PAT on the workflow, and it's just using the built-in github token, no CI is being triggered right now on release PRs. You can look at the current release PR to see that. This gives you a way to trigger CI on the release PRs (for a sanity check) if you want. I summarized the issue here: https://github.com/eslint-community/eslint-utils/pull/294#issuecomment-3707361402

Let me know what you think

aladdin-add: can we use the solution as in https://github.com/eslint/css/pull/330/?

michaelfaith: Possibly, but that requires onboarding the repo to an app that I don't believe it's currently using, and i don't have the ability to set that up (requires repo setting changes) .  This is just simpler 🤷‍♂️. As a point of reference, this is how `yargs` is doing it.

aladdin-add: Since the CI for this repository is quite time-consuming (with a large number of tasks), I'm a bit concerned that it might trigger some unnecessary CI checks. Let's focus on the release-please configuration in this PR. 

We can create a new PR once the discussion at https://github.com/eslint-community/eslint-utils/pull/294#issuecomment-3707361402 reaches a conclusion. wdyt?

michaelfaith: Sounds good to me. I'll update

michaelfaith: Updated

aladdin-add (approved):
LGTM, thanks! ❤️

PR #578, opened 2026-01-04
fix: address ci failures

### LINKED ISSUE #575: Bug: CI has been failing since November 10th
@aladdin-add looks like ever since this merge in November, type checking has been failing in main https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/19239812032

cc @StyleShit

michaelfaith:
This change addresses the two different kind of type check failures:
- 2 rules weren't using explicit types when they were needed
- `eslint-scope`'s `ScopeManager` type is incompatible with `eslint`'s built-in `Scope.Scopemanager`

Normally I would have solved this by using `patch-package` to patch the `@types/eslint-scope` package directly, but `patch-package` requires a lock file in the repo 😢.  So, I just forked the types entirely, and replace `ScopeManager` with `eslint`'s `Scope.ScopeManager`.

Closes #575

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,8 +62,8 @@
     "@types/estree": "^1.0.8",
     "@types/lodash": "^4.17.18",
     "@types/node": "^20.19.23",
-    "@typescript-eslint/parser": "^8.51.0",
-    "@typescript-eslint/utils": "^8.51.0",
+    "@typescript-eslint/parser": "8.47.0",
+    "@typescript-eslint/utils": "8.47.0",
     "@vitest/coverage-istanbul": "^3.2.4",
     "eslint": "^9.39.1",
     "eslint-config-not-an-aardvark": "^2.1.0",
@@ -85,7 +85,7 @@
     "prettier": "3.6.2",
     "tsdown": "^0.13.3",
     "typescript": "^5.9.2",
-    "typescript-eslint": "^8.51.0",
+    "typescript-eslint": "8.47.0",
     "vitest": "^3.2.4"
   },
   "peerDependencies": {
--- a/lib/rules/no-property-in-node.ts
+++ b/lib/rules/no-property-in-node.ts
@@ -40,8 +40,6 @@ function isAstNodeType(
     )
     .some((declaration) => {
       const fileName = declaration.getSourceFile().fileName;
-      // eslint-disable-next-line no-console -- TEMPORARY: Debug logging
-      console.log('AST node type declaration source file:', fileName);
       return (
         fileName &&
         typedNodeSourceFileTesters.some((tester) => tester.test(fileName))
--- a/tests/lib/rules/no-property-in-node.ts
+++ b/tests/lib/rules/no-property-in-node.ts
@@ -5,8 +5,6 @@ import rule from '../../../lib/rules/no-property-in-node.ts';
 import parser from '@typescript-eslint/parser';
 
 const dirname = path.dirname(fileURLToPath(import.meta.url));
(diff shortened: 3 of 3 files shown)


michaelfaith:
No, still failing.  I'm not sure why I can't reproduce the failures locally. 

image

michaelfaith:
I just pushed a console.log on the filename to see what's going through that code in CI.

michaelfaith:
Yeah, so, on CI it's not resolving to the type files.  

This is what CI is reporting for file names

image


This is why my local (and probably yours) is reporting.  The ones where it's @types/estree or @typescript-eslint those are successful test evaluations.

image

michaelfaith:
This is the first point where these 4 tests started failing in CI: https://github.com/eslint-community/eslint-plugin-eslint-plugin/actions/runs/19694297600/job/56416156941  (November 26th)

package.json
bmish, line 1: If adding a lockfile makes things easier, I'm not opposed to adding one. We just didn't need one in the past. But no strong reason to avoid it.

MichaelDeBoey:
Sound like a solid plan @michaelfaith 👍

michaelfaith:
Looking in the `typescript-eslint` code around the `getTypeAtLocation()` function, it looks as though, in `createParserService` it's setting up the parser service to delegate to `typescript` for this function, but using a custom ast map to pass in.  

The `createParserService` code hasn't changed in a year.  But the `./convert` code where the ast map that's used to pass into typescript's `getTypeAtLocation` _has_ changed, and there were specifically two changes to it that happened in the 8.48.0 release.

https://github.com/typescript-eslint/typescript-eslint/blame/cf79108b6405972fb73f5991e913e1b36de8a67f/packages/typescript-estree/src/createParserServices.ts#L35-L36

image

The first one, is in the part of the code that deals with `SyntaxKind.ImportType`.  I wonder if that could be at play?  The second change only does a type cast (`as`).

michaelfaith:
Opened https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/579 and pinned typescript-eslint

StyleShit (approved):

MichaelDeBoey (approved):

MichaelDeBoey:
As @fasttime mentioned in https://github.com/eslint/js/issues/719#issuecomment-3709348758, it seems the ESLint team isn't planning on adding a fix for ESLint v9, so we should probably fix [`@types/eslint-scope`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/eslint-scope) ourselves

> There are currently no plans to backport the changes in [#708](https://github.com/eslint/js/issues/708) to `eslint-scope` v8 (used by ESLint v9). Meanwhile, if there's an issue with `@types/eslint-scope` it can be addressed by submitting a PR at [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), which is an external repository not maintained by the ESLint team.

We can probably take a few things from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74066 for that

MichaelDeBoey:
> but they've left the v9 types in a bad state?

I think the idea is "we never were responsible for this ourselves, so we don't/shouldn't care now either"
I personally also don't agree with the fact that they leave ESLint v9 users using TypeScript in a bad state, but @fasttime can probably give more insights into the thought process of this.

> Sure `@types/eslint-scope` is a different repo, but if the argument of not fixing it in the DT repo was that they'd be absorbed into the core package, I would expect the DT types to be supported until that happens and not leave users in limbo like that.

I have the exact same feeling

PR #620, opened 2026-05-28
feat: add new rule `require-test-error-positions`

### LINKED ISSUE #59: Rule: consistent RuleTester.error properties
When writing ESLint rule tests using [RuleTester](https://eslint.org/docs/developer-guide/nodejs-api#ruletester), it can be beneficial to enforce consistent usage of the `errors` property in this RuleTester docs screenshot:

image

For example, I usually always want to enforce `errors` to be an object, and I want to require that object to include the `message`, `line`, and `column` properties. 

Thoughts on adding a rule like this? Still thinking through what the name could be and what the default options would be for this rule.

### LINKED ISSUE #485: New rule `require-positions`
Many contributors add tests with only `line: 4` for reported errors. Often, the test cases should include the column and end position though. This could be enforced with a new ESLint rule that requires all of `line`, `endLine`, `column`, and `endColumn` to be present for all expected errors.

Vinccool96:

Adds a new rule to require the position of errors to be explicitly stated for all expected errors.

* Added the rule
* Disabled it in the eslint config while waiting for approval of the design

#### Related Issues

fixes #485 
fixes #59

Currently, I need to know if my design is correct. If it is, I will reactivate it in the eslint config and add fix all its occurrences in the repo.


docs/rules/consistent-test-errors.md
bmish, line 1: We should link to the API docs which list all of the possible properties:
- https://eslint.org/docs/latest/integrate/nodejs-api#ruletesterrun

docs/rules/consistent-test-errors.md
```diff
@@ -0,0 +1,67 @@
+# eslint-plugin/consistent-test-errors
+
+📝 Enforces consistent definition of all expected errors in rule tests.
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/eslint-community/eslint-plugin-eslint-plugin#presets).
+
+<!-- end auto-generated rule header -->
+
+Requires all of `line`, `endLine`, `column`, and `endColumn` to be present for all expected errors.
```
bmish: @Vinccool96 apologies for the churn. Can we actually revert / rename it back to `require-error-positions`? On second thought, I don't think we need a generic rule to support any arbitrary error object shape, as I realized that ESLint already [enforces](https://eslint.org/docs/latest/integrate/nodejs-api#ruletesterrun) the other properties when they are necessary (some of this from the [RFC](https://github.com/eslint/rfcs/tree/main/designs/2021-stricter-rule-test-validation) I implemented a while back).

So I think it would actually be better to keep this rule focused on requiring the optional position properties. This wouldn't be a recommended rule, but would be available for anyone interested in this enforcement.

docs/rules/consistent-test-errors.md
```diff
@@ -0,0 +1,67 @@
...
+});
+```
+
+## Options
+
+<!-- begin auto-generated rule options list -->
+
+| Name              | Description                                                               | Type    | Default |
+| :---------------- | :------------------------------------------------------------------------ | :------ | :------ |
+| `requireLocation` | Whether to enforce stating the position of errors to be explicitly stated | Boolean | `true`  |
```
bmish: Don't think we need any options after we rename it to `require-error-positions`.

bmish (changes requested):

docs/rules/consistent-test-errors.md
Vinccool96: Do we still link to the possible properties if we revert back to `require-error-positions`?

eslint.config.ts
```diff
@@ -74,6 +74,8 @@ export default defineConfig([
             'https://github.com/eslint-community/eslint-plugin-eslint-plugin/tree/HEAD/docs/rules/{{name}}.md',
         },
       ],
+      // TODO reactivate when it is approved
+      'eslint-plugin/require-error-positions': 'off',
```
bmish: Perhaps we can just omit this? It seems like it would be a lot of work to enable it. If we really wanted to enable it, could be done in a separate PR.

Vinccool96: So just remove the TODO comment?

bmish: It's off by default so we can remove both of these lines.
```suggestion
```

lib/rules/require-error-positions.ts
```diff
@@ -0,0 +1,109 @@
+import type { Rule } from 'eslint';
+import type { ObjectExpression, Property } from 'estree';
+
+import { getTestInfo } from '../utils.ts';
+
+const rule: Rule.RuleModule = {
+  meta: {
+    type: 'problem',
```
bmish: This rule is a suggestion. It doesn't indicate an actual problem or bug.

```suggestion
    type: 'suggestion',
```

https://eslint.org/docs/latest/extend/custom-rules#rule-structure

eslint.config.ts
Vinccool96: I can assure you that removing those lines makes the rule be enabled by default, because of these lines https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/9ddb181bdbf1a65f8c243528b7b0c3da7c853ee9/eslint.config.ts#L65-L67

lib/rules/require-error-positions.ts
```diff
@@ -0,0 +1,109 @@
...
+    docs: {
+      description:
+        'requires the position of errors to be explicitly stated for all expected errors',
+      category: 'Tests',
+      recommended: true,
```
bmish: I don't think this rule is suitable to become an official `recommended` because:
- Not all test cases need to be this explicit about where exactly the error is--it can be overkill to require 4 additional properties in every test case
- It can be tedious to add these properties to test cases (and there's no autofixer to add them)
- It's more of a rule author preference in how tests should be written, not everyone wants or needs to these properties

```suggestion
      recommended: false,
```

That said, it's fine to be an optional rule for anyone who would like to opt-in to it.

(And note, adding a `recommended` rule is a breaking change anyway, which we only do during major releases).

eslint.config.ts
bmish: Thanks, you're right that we enable all the rules. Let's keep it disabled for now. You can follow-up in a separate PR to enable it for this plugin if you would like to.

docs/rules/consistent-test-errors.md
bmish: It's good in general to link to any relevant document / API spec.

lib/rules/require-error-positions.ts
```diff
@@ -0,0 +1,109 @@
...
+    fixable: undefined,
+    schema: [],
+    messages: {
+      locationsMissing:
+        'The full location of the error must be specified using the properties `column`, `endColumn`, `endLine`, and `line`',
+    },
+  },
+
+  create(context) {
+    const locationProperties: readonly string[] = [
```
bmish: > Redundant typing. const locationProperties: readonly string[] = [...] as const; — the explicit readonly string[] annotation widens the as const tuple back to string[], so the as const buys nothing here. Drop as const.

lib/rules/require-error-positions.ts
```diff
@@ -0,0 +1,109 @@
...
+        getTestInfo(context, ast).forEach((testRun) => {
+          testRun.invalid
+            .filter((test) => !!test)
+            .forEach((test) => {
+              /**
+               * Get a test case's given key name node.
+               * @param key the keyname to find.
+               * @returns found node; if not found, return null;
+               */
+              function getTestInfoProperty(key: string): Property | null {
```
bmish: > Duplicated helper. getTestInfoProperty is copy-pasted verbatim from prefer-output-null.ts (comment typo and all — the original's @param the keyname is at least preserved). Shared util would be nice.

tests/lib/rules/require-test-error-positions.ts
bmish, line 1: > Gaps
> The test file covers the basic valid/invalid happy path but is thin. I'd add cases for:
> - String-literal keys (`'line': 1, ...`) — currently a false positive (issue #1).
> - Partial positions present (e.g. only `line` + `column`) — confirms it requires all four, not just "any".
> - Multiple error objects in one array where some are valid and some aren't.
> - A spread `{ ...errorPositions }` case — the docs claim this is incorrect, but there's no test asserting it.
> - String-form error element (issue #2), once behavior is decided.
>

lib/rules/require-error-positions.ts
```diff
@@ -0,0 +1,109 @@
...
+      'line',
+    ] as const;
+
+    function verifyErrorLocations(error: ObjectExpression) {
+      const existingLocationProperties = new Set<string>();
+      const properties = error.properties.filter(
+        (property) => property.type === 'Property',
+      );
+
+      properties.forEach((property) => {
```
bmish: > 
> 
> **1. String-literal and computed keys cause false positives.** The rule detects properties manually:
> 
> ```48:60:lib/rules/require-error-positions.ts
>       properties.forEach((property) => {
>         if (property.key.type !== 'Identifier') {
>           return;
>         }
> 
>         if (locationProperties.includes(property.key.name)) {
>           existingLocationProperties.add(property.key.name);
>         }
>       });
> ```
> 
> This only matches `Identifier` keys, so a perfectly valid test case written as `{ messageId: 'x', 'line': 1, 'column': 2, 'endColumn': 3, 'endLine': 4 }` (string-literal keys) gets falsely reported. The rest of this codebase uses the `getKeyName()` helper from `lib/utils.ts` precisely to normalize `Identifier`, `Literal`, and computed keys (`consistent-output.ts` and `getTestInfo` itself rely on it). This rule should do the same instead of hand-rolling the `Identifier` check.
>

lib/rules/require-test-error-positions.ts
```diff
@@ -0,0 +1,109 @@
...
+              }
+
+              const errors = errorsProperty.value;
+
+              // Only verify the errors that are ArrayExpression
+              if (errors.type !== 'ArrayExpression') {
+                return;
+              }
+
+              errors.elements
```
bmish, line 92: > String error elements silently escape the rule.** The rule only inspects `ObjectExpression` error elements:
> 
> ```92:99:lib/rules/require-error-positions.ts
>               errors.elements
>                 .filter(
>                   (element) => !!element && element.type === 'ObjectExpression',
>                 )
>                 .forEach((element) => {
>                   verifyErrorLocations(element);
>                 });
> ```
> 
> An error written as a bare string (`errors: ['some message']`) has no position properties at all, yet it's skipped rather than flagged. That seems contrary to the rule's intent ("require positions for all expected errors"). Either it should report string-form errors as missing positions, or the docs should explicitly call out that string-form errors are intentionally exempt. Worth a decision + a test either way.
>

eslint.config.ts
Vinccool96: I will indeed enable it and resolve it in a subsequent PR

docs/rules/require-error-positions.md
```diff
@@ -0,0 +1,51 @@
...
+
+```js
+/* eslint eslint-plugin/require-error-positions: error */
+
+const errorPositions = { line: 1, column: 2, endColumn: 3, endLine: 4 };
+
+new RuleTester().run('foo', bar, {
+  valid: [],
+  invalid: [{ code: 'foo', errors: [{ messageId: 'bar', ...errorPositions }] }],
+});
```
bmish: This case looks like it should be valid. We can use `getStaticValue` for it, as other rules do.

```ts
const scope = context.sourceCode.getScope(error);

for (const property of error.properties) {
  if (property.type === 'Property') {
    const key = getKeyName(property);
    if (key) existingLocationProperties.add(key);
  } else {
    // SpreadElement — resolve `...errorPositions` to its static object value
    const staticValue = getStaticValue(property.argument, scope);
    if (staticValue && typeof staticValue.value === 'object' && staticValue.value) {
      for (const key of Object.keys(staticValue.value)) {
        existingLocationProperties.add(key);
      }
    }
  }
}
```

docs/rules/require-error-positions.md
```diff
@@ -0,0 +1,51 @@
...
+
+📝 Requires the position of errors to be explicitly stated for all expected errors.
+
+<!-- end auto-generated rule header -->
+
+Requires all the [position properties](https://eslint.org/docs/latest/integrate/nodejs-api#ruletesterrun) (`line`, `endLine`, `column`, and `endColumn`) to be present for all expected errors.
+
+## Rule Details
+
+The rule reports that doesn't have consistent test errors definition
```
bmish:
```suggestion
This rule reports any expected error in an invalid test case that does not explicitly specify all four position properties.
```

lib/index.ts
```diff
@@ -110,6 +111,7 @@ const allRules = {
   'prefer-placeholders': preferPlaceholders,
   'prefer-replace-text': preferReplaceText,
   'report-message-format': reportMessageFormat,
+  'require-error-positions': requireErrorPositions,
```
bmish: Would you mind renaming this rule one more time to `require-test-error-positions`? This makes it consistent with `require-test-case-name`, and makes it clear it's referring to tests rather than error reporting inside rules. Apologies again for the churn. After that, I am ready to approve and merge the PR.

docs/rules/require-test-error-positions.md
```diff
@@ -0,0 +1,62 @@
...
+Requires all the [position properties](https://eslint.org/docs/latest/integrate/nodejs-api#ruletesterrun) (`line`, `endLine`, `column`, and `endColumn`) to be present for all expected errors.
+
+## Rule Details
+
+This rule reports any expected error in an invalid test case that does not explicitly specify all four position properties.
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/require-error-positions: error */
```
bmish:
```suggestion
/* eslint eslint-plugin/require-test-error-positions: error */
```

docs/rules/require-test-error-positions.md
```diff
@@ -0,0 +1,62 @@
...
+
+new RuleTester().run('foo', bar, {
+  valid: [],
+  invalid: [{ code: 'foo', errors: [{ messageId: 'bar' }] }],
+});
+```
+
+```js
+/* eslint eslint-plugin/require-error-positions: error */
```
bmish:
```suggestion
/* eslint eslint-plugin/require-test-error-positions: error */
```

docs/rules/require-test-error-positions.md
```diff
@@ -0,0 +1,62 @@
...
+new RuleTester().run('foo', bar, {
+  valid: [],
+  invalid: [{ code: 'foo', errors: [{ messageId: 'bar', ...errorPositions }] }],
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+/* eslint eslint-plugin/require-error-positions: error */
```
bmish:
```suggestion
/* eslint eslint-plugin/require-test-error-positions: error */
```

docs/rules/require-test-error-positions.md
```diff
@@ -0,0 +1,62 @@
...
+      errors: [
+        { messageId: 'bar', line: 1, column: 2, endColumn: 3, endLine: 4 },
+      ],
+    },
+  ],
+});
+```
+
+```js
+/* eslint eslint-plugin/require-error-positions: error */
```
bmish:
```suggestion
/* eslint eslint-plugin/require-test-error-positions: error */
```

bmish (approved):

bmish:
Nice work!

PR #624, opened 2026-06-01
chore: enabled and fixed `require-test-error-positions` errors in project

### LINKED ISSUE #622: Chore: Enable `require-test-error-positions` in the repository
With the new rule `require-test-error-positions` added by the PR #620 , the new rule was disabled since it would have required to fix the tests in pretty much every test files. Thus, the rule must now be enabled, and the test files must be fixed.

Vinccool96:

This PR enables and resolves the new rule `require-test-error-positions` in the project

* Enabled `require-test-error-positions` in the eslint config
* Fixed the rule in all the offending files

#### Related Issues

Fixes #622 

🥰


tests/lib/rules/fixer-return.ts
bmish, line 1: Was this all done manually? Or AI presumably? I guess AI is the best way to autofix this rule and there's no other automated way to do it?

tests/lib/rules/fixer-return.ts
```diff
@@ -291,8 +291,10 @@ ruleTester.run('fixer-return', rule, {
         {
           messageId: 'missingFix',
           type: 'FunctionExpression',
-          line: 5,
           column: 24,
+          endColumn: undefined,
+          endLine: undefined,
```
bmish, line 296: What is the meaning of `undefined` here?

Vinccool96: Apparently, this rule doesn't have an end location, only a start location.

tests/lib/rules/fixer-return.ts
Vinccool96: Manually. I don't know how I could have guessed the error locations beforehand.

bmish (approved):

PR #631, opened 2026-06-29
fix: Avoid false positives with unresolved meta object spreads in `require-meta-*` rules

### LINKED ISSUE #274: False positives when meta contains spread with `prefer-message-ids`, `require-meta-type`, `require-meta-fixable`
When there's a spread inside the meta object that we can't evaluate, we might want to stop flagging these violations since likely the spread fills these properties in already:

```
  23:9  error  `meta.messages` must contain at least one violation message                    eslint-plugin/prefer-message-ids
  23:9  error  `meta.type` is required (must be either `problem`, `suggestion`, or `layout`)  eslint-plugin/require-meta-type
  23:9  error  `meta.fixable` must be either `code` or `whitespace` for fixable rules         eslint-plugin/require-meta-fixable
```

```ts
export default createRule<Options, MessageIds>({
  name: 'object-curly-spacing',
  meta: {
    ...baseRule.meta,
    docs: { /* ... * / }
  }
});
```

Example: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/object-curly-spacing.ts

morgan-coded:
The `require-meta-*` rules could report missing meta properties when a rule used an unresolved spread like `meta: { ...baseRule.meta }`.

This suppresses the missing-property report for the affected meta rules only when the property is absent and the meta object contains an unresolved object spread. I checked the negative control with the expected false positives, then ran the full suite with 1,186 tests passing and coverage thresholds met.

Fixes #274

BEGIN_COMMIT_OVERRIDE
fix: Avoid false positives with unresolved meta object spreads in `require-meta-*` rules
END_COMMIT_OVERRIDE

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -920,6 +920,57 @@ export function isSuggestionFixerFunction(
   );
 }
 
+// Spreads of these node types contribute no statically-known named property that could match a
+// rule-meta key (arrays and strings expose only numeric indices; primitives, functions, and
+// classes have no object-literal members to read). They are treated as resolved-but-empty — so a
+// genuinely missing meta key is still reported — rather than as an unknown (suppressed) spread.
+const SPREAD_TYPES_WITHOUT_NAMED_PROPERTIES = new Set([
+  'ArrayExpression',
+  'ArrowFunctionExpression',
+  'ClassExpression',
+  'FunctionDeclaration',
+  'FunctionExpression',
+  'Literal',
+  'TemplateLiteral',
+]);
+
+type SpreadResolution =
+  | { kind: 'object'; node: ObjectExpression }
+  | { kind: 'empty' }
+  | { kind: 'unknown' };
+
+/**
+ * Classify the value that a spread element spreads, resolving an identifier one hop.
+ * - `object`: spreads an object expression (inline `...{ a: 1 }` or a variable initialized
+ *   to one) whose properties can be inspected and recursed into.
+ * - `empty`: spreads a statically-known value that cannot contribute a named property
+ *   (e.g. `...{}`, `...5`, `...[1, 2]`, `...() => {}`).
+ * - `unknown`: spreads a value whose properties cannot be determined statically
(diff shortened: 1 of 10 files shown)


lib/utils.ts
```diff
@@ -950,6 +950,27 @@ export function evaluateObjectProperties(
   });
 }
 
+export function hasUnresolvedObjectSpread(
```
bmish, line 1067: > `hasUnresolvedObjectSpread()` should recurse into resolved object spreads here. Right now an identifier spread is treated as resolved whenever `findVariableValue()` returns an `ObjectExpression`, but that object expression may itself contain an unresolved spread that `evaluateObjectProperties()` will later drop.
> 
> For example:
> 
> ```ts
> const inheritedMeta = { ...baseRule.meta };
> 
> module.exports = {
>   meta: { ...inheritedMeta },
>   create(context) {}
> };
> ```
> 
> In this case `hasUnresolvedObjectSpread(meta, scopeManager)` returns `false` because `inheritedMeta` resolves to an object expression. But `evaluateObjectProperties(meta, scopeManager)` expands `inheritedMeta`, then discards the nested `...baseRule.meta`, so the rule still reports missing `meta.type` / `meta.schema` / `meta.messages` / `meta.fixable`. That leaves the original false positive unresolved for extracted metadata helpers.
> 
> A fix could make this helper follow resolved object-expression spreads recursively: if the spread argument is an object expression, inspect it; if it is an identifier resolving to an object expression, inspect that object; otherwise treat it as unresolved. That would also pair well with the existing inline-spread false-negative comment, since `...{}` would be recognized as resolved instead of automatically suppressing checks.
>

morgan-coded: Thanks, that makes sense. I updated the helper to recurse through resolved object spreads, and to only suppress the missing-meta checks when the spread is still genuinely unresolved.

bmish (approved):
Thanks! I extracted #633 as a separate issue.

PR #639, opened 2026-07-11
fix: Avoid false positives for external meta docs in `require-meta-*` rules

### LINKED ISSUE #633: Bugs with `docs` spreads
Extracting this from other issue:

> I have similar issues with `eslint-plugin/require-meta-docs-*` rules. Basicually if I move the `docs` definition out of the `meta` JSON, I cannot get any of the rules to pass e.g. both would fail:
> ```js
> const docs = require("./rule.docs.js");
> meta: {
>     docs
> } 
> ```
> and
> ```js
> const docs = require("./rule.docs.js");
> meta: {
>    docs: { ...docs }
> }
> ...
> 
> At the moment I work this around by manually adding every single property to the `docs` array. 

 _Originally posted by @tonyganchev in [#274](https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/274#issuecomment-3448134974)_

morgan-coded:
The `require-meta-docs-*` rules could report missing properties when `meta.docs` came from an imported module or spread a required docs object, even though those properties could not be inspected locally. This leaves known local objects checked as before while skipping missing-property reports for external-module-backed docs. Checked it with the three focused rule suites, the full 1,210-test suite with coverage, lint, typecheck, and build. Fixes #633

This also resolves local `meta: { docs }` references through `resolveSpreadObject()` before looking up the requested docs property, which fixes the false positive on main when `docs` is a local object constant. The review revision now reuses the existing unresolved-spread machinery to detect when `meta.docs` may contain unknown properties.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -971,61 +971,6 @@ function resolveSpreadObject(
   return { kind: 'unknown' };
 }
 
-function objectValueMayComeFromExternalModule(
-  value: Expression,
-  scopeManager: Scope.ScopeManager,
-  visited = new Set<Node>(),
-): boolean {
-  if (visited.has(value)) {
-    return false;
-  }
-  visited.add(value);
-
-  if (value.type === 'Identifier') {
-    const variable = findVariable(
-      scopeManager.acquire(value) || scopeManager.globalScope!,
-      value,
-    );
-    if (
-      variable?.defs.some((definition) => definition.type === 'ImportBinding')
-    ) {
-      return true;
-    }
-    const resolvedValue = findVariableValue(value, scopeManager);
-    return resolvedValue?.type === 'FunctionDeclaration'
-      ? false
-      : Boolean(
-          resolvedValue &&
-          objectValueMayComeFromExternalModule(
-            resolvedValue,
-            scopeManager,
-            visited,
-          ),
-        );
-  }
-
-  if (
-    value.type === 'CallExpression' &&
-    value.callee.type === 'Identifier' &&
-    value.callee.name === 'require'
-  ) {
-    return true;
-  }
-
-  return (
-    value.type === 'ObjectExpression' &&
-    value.properties.some(
-      (property) =>
-        property.type === 'SpreadElement' &&
-        objectValueMayComeFromExternalModule(
-          property.argument,
-          scopeManager,
-          visited,
-        ),
-    )
-  );
-}
-
 /**
  * List all properties contained in an object.
  * Evaluates and includes any properties that may be behind spreads.
@@ -1193,17 +1138,15 @@ export function getMetaDocsProperty(
     .filter((node) => node.type === 'Property')
     .find((p) => getKeyName(p) === propertyName);
 
-  const docsPropertyMayExist = Boolean(
+  const docsMayHaveUnknownProperties = Boolean(
     docsNode &&
-    objectValueMayComeFromExternalModule(
-      docsNode.value as Expression,
-      scopeManager,
-    ),
+    (docsResolution?.kind === 'unknown' ||
+      hasUnresolvedObjectSpread(docsObjectNode, scopeManager)),
   );
 
   return {
     docsNode,
-    docsPropertyMayExist,
+    docsMayHaveUnknownProperties,
     metaNode,
     metaPropertyNode,
   };
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -971,6 +971,61 @@ function resolveSpreadObject(
   return { kind: 'unknown' };
 }
 
+function objectValueMayComeFromExternalModule(
+  value: Expression,
+  scopeManager: Scope.ScopeManager,
+  visited = new Set<Node>(),
+): boolean {
+  if (visited.has(value)) {
+    return false;
+  }
+  visited.add(value);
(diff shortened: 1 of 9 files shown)


lib/utils.ts
bmish, line 1: The AI review makes some good points including about reusing some existing util functions.

> I have one main request before this lands, plus a few smaller notes.
> 
> Reuse the existing unresolved-spread machinery instead of adding a new util
> 
> We already solved this exact problem for require-meta-type, require-meta-schema, require-meta-fixable, and prefer-message-ids via hasUnresolvedObjectSpread() / resolveSpreadObject()'s 'unknown' classification. The new objectValueMayComeFromExternalModule() reimplements a narrower version of that ("comes from an import/require" rather than "not statically resolvable"), so these closely related shapes still false-positive on this branch (I verified against the built rules):
> 
> ```ts
> // all still report "missing":
> meta: { docs: baseRule.meta.docs }
> meta: { docs: { ...baseRule.meta.docs } }  // the extend-a-base-rule pattern from #274
> meta: { docs: getDocs() }
> ```
> 
> Since getMetaDocsProperty() already computes docsResolution, the new function can be replaced with something like:
> 
> ```ts
> const docsPropertyMayExist = Boolean(
>   docsNode &&
>     (docsResolution?.kind === 'unknown' ||
>       hasUnresolvedObjectSpread(docsObjectNode, scopeManager)),
> );
> ```
> 
> This covers everything the PR covers (a require() call resolves to 'unknown'; an import binding has no resolvable value, so also 'unknown'; spreads go through hasUnresolvedObjectSpread), fixes the member-expression/call-expression cases above, keeps docs: {} correctly reporting, and keeps these three rules consistent with the four rules that already handle unresolved spreads. It also drops ~55 lines of bespoke recursion. Test cases for the shapes above would be great to add alongside.
> 
> The getMetaDocsProperty() change fixes a second bug — please test it
> 
> Resolving docsNode.value through resolveSpreadObject() before the property lookup means this is now correctly valid (it false-positived on main):
> 
> const docs = { description: 'disallow foo' };
> module.exports = { meta: { docs }, create(context) {} };
> 
> That's a genuinely useful fix riding along — worth a mention in the PR description and a couple of tests so it doesn't regress:
> 
> - valid: local const docs = { description: '...' } referenced as meta: { docs }
> - invalid: local const docs = {}; referenced as meta: { docs } (should still report)
> 
> Smaller notes
> 
> - Naming: docsPropertyMayExist reads as "the docs property may exist," but it's only true when docsNode was found — it really means "the requested property inside docs may exist but can't be verified." Something like docsMayHaveUnknownProperties would read better at the call sites.
> - Unit tests: tests/lib/utils.ts tests hasUnresolvedObjectSpread directly; the getMetaDocsProperty changes are only covered indirectly through rule tests. A few direct cases there would be welcome. (Moot for the new util if it's removed per the above.)
> - Out of scope but adjacent: const meta = require('./meta.js'); module.exports = { meta, create } still reports missing since there's no docsNode at all. Fine to defer — happy to track it in a follow-up issue since it's the same root cause one level up.
> 
> The trade-off that external docs now silently pass even if the external file genuinely lacks the property is fine — it matches the precedent of the other rules and is inherent to static analysis.

bmish: Follow-up:

> Thanks for the update — this addressed everything from the first round nicely. The reuse of resolveSpreadObject/hasUnresolvedObjectSpread keeps these rules consistent with the other meta rules, the base-rule/member-expression/call-expression shapes are all covered, and the getMetaDocsProperty unit suite is a great addition. I verified the full test suite, lint, and typecheck locally, and confirmed the previously false-positive patterns now pass while docs: {} still reports.
> 
> One remaining issue I'd like fixed in this PR, since it's a small regression the new resolution introduces:
> 
> meta: { docs: undefined } is now silently skipped instead of reported.
> 
> On main this reports missing; on this branch, undefined is an identifier whose global variable has no resolvable definition, so resolveSpreadObject() classifies it as 'unknown' and the rules bail out. That's inconsistent in two ways:
> 
> - docs: null still reports missing (it's a Literal, so it's classified as 'empty') — only undefined slips through.
> - Everywhere else in this plugin, an explicit undefined is treated as statically known and reportable: description: undefined → wrongType, recommended: undefined → incorrect (with a dedicated isUndefinedIdentifier path), type: undefined → unexpected, schema: undefined → wrongType, and the require-meta-docs-url fixer even special-cases replacing an undefined identifier.
> 
> The dividing line the codebase already draws is "skip what you can't know, report what you can" — and undefined is knowable. The clean fix is in resolveSpreadObject(), before the findVariableValue resolution:
> 
> if (node.type === 'Identifier' && isUndefinedIdentifier(node)) {
>   return { kind: 'empty' };
> }
> 
> That's also semantically correct for spreads ({ ...undefined } is legal JS that contributes no properties), so it's a strict improvement for all six rules that use this util, not a docs-rules special case.
> 
> Could you add that along with:
> - a docs: undefined case in the new getMetaDocsProperty unit suite (docsMayHaveUnknownProperties: false, no property found), and
> - one invalid rule test (e.g. require-meta-docs-description reporting missing for meta: { docs: undefined })?

bmish (approved):
Thanks!

Issue #538, opened 2025-08-04
V8 Planned Changes


bmish:
Starting a list of breaking changes, feedback welcome. We can perform this release once we have compiled enough changes.

Planned:
* Enable additional `recommended` rules:
   * https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/563
   * TODO
* Change default rule options
   * (TODO: decide on this) https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/495 
   * (TODO: decide on this) https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/475 
* Supported Node.js versions
  * 2026-04-30: Node 20 end-of-life ([schedule](https://github.com/nodejs/Release?tab=readme-ov-file#release-schedule))
  * TODO
* Supported ESLint versions
  * TODO
* Migrations
  * TODO

For reference, V7 was released 2025-08-04 ([release changelog](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/releases/tag/v7.0.0), discussion in #505).

michaelfaith:
For consideration https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/644 (adding `meta.languages` to all rules)

PR #643, opened 2026-07-15
feat: add new rule `require-meta-languages`

### LINKED ISSUE #614: New rule `require-meta-languages`
Eslint v10.2.0 introduced `meta.languages`. 

https://eslint.org/blog/2026/04/eslint-v10.2.0-released/

It would be helpful to have a new rule to require the meta language be set for rules.

This new rule might also be worth considering for the recommended config, if not immediately then in the next major release.

### LINKED ISSUE #645: Feature: validate language-identifier format and duplicates in `require-meta-languages`
Following up on @bmish's review of #643: the `require-meta-languages` rule currently accepts any non-empty array of strings for `meta.languages` without inspecting the individual entries, so both of these pass today:

```js
// not a real language identifier
module.exports = { meta: { languages: ['javascript'] }, create(context) {} };

// duplicate entry
module.exports = { meta: { languages: ['js/js', 'js/js'] }, create(context) {} };
```

A future enhancement could:

- report entries that don't match the `pluginNamespace/languageName` identifier shape (e.g. `'js/js'`, `'json/json'`); and
- flag duplicate entries.

Tracking this separately from #643 as suggested. Since `meta.languages` is still fairly new and its semantics may evolve, it's worth settling how strictly to validate the identifier format — and against what source of truth — before implementing.

This would be a non-breaking addition to the existing (non-recommended) rule.

morgan-coded:
**What changed:** Adds a new `require-meta-languages` rule for the `meta.languages` property introduced in ESLint v10.2.0, reporting absent properties, empty arrays, and values that are not arrays of strings while skipping dynamic values and unresolved `meta` spreads that cannot be inspected statically.

Registers the rule in `lib/index.ts` and adds its rule doc plus the README row generated by `eslint-doc-generator`; it stays out of `recommended` because enabling it there is a breaking change deferred to #538.

**ESLint 9 compatibility:** As agreed on #614, keeps the rule off for this plugin's own files in `eslint.config.ts` because the plugin still supports ESLint 9, whose rule metadata types do not include `meta.languages`; it can be enabled there once the minimum ESLint version reaches 10.2.

**Testing:** Unit tests cover absent, empty, invalid, unresolved, and valid values, and the full suite passes with 1,244 tests and 98.3% statement, 96.05% branch, 99.17% function, and 98.44% line coverage.

Manual validation also ran the built rule against `eslint-plugin-n` 18.2.2, `eslint-plugin-unicorn` 72.0.0, and `eslint-plugin-promise` 7.3.0; all three runs were crash-free, the unresolved `meta` spread in `eslint-plugin-n` was skipped, and Unicorn rules already declaring `languages: ['js/js']` passed.

Closes #614

Closes #645

### DIFF
--- a/docs/rules/require-meta-languages.md
+++ b/docs/rules/require-meta-languages.md
@@ -1,0 +1,35 @@
+# eslint-plugin/require-meta-languages
+
+📝 Require rules to implement a `meta.languages` property.
+
+<!-- end auto-generated rule header -->
+
+## Rule Details
+
+This rule requires ESLint rules to have a non-empty `meta.languages` array containing only strings.
+
+Examples of **incorrect** code for this rule:
+
+```js
+/* eslint eslint-plugin/require-meta-languages: error */
+
+module.exports = {
+  meta: {},
+  create(context) {
+    /* ... */
+  },
+};
+```
+
+Examples of **correct** code for this rule:
+
+```js
+/* eslint eslint-plugin/require-meta-languages: error */
+
+module.exports = {
+  meta: { languages: ['js/js'] },
+  create(context) {
+    /* ... */
+  },
+};
+```
--- a/docs/rules/require-meta-languages.md
+++ b/docs/rules/require-meta-languages.md
@@ -4,6 +4,12 @@
 
 <!-- end auto-generated rule header -->
 
+ESLint v10.2.0 introduced `meta.languages` to declare which languages a rule supports, using plugin-prefixed identifiers such as `'js/js'` or `'json/json'`.
+
+When `meta.languages` is omitted, ESLint assumes the rule supports every language. This assumption is incorrect for almost all rules: a rule written for JavaScript can crash on an unfamiliar JSON or CSS AST, or silently do nothing. Declaring `meta.languages` allows ESLint to report a clear configuration-time error instead.
+
+Dynamically computed values and unresolved spreads in `meta` are skipped, matching the behavior of the other `require-meta-*` rules.
+
 ## Rule Details
 
 This rule requires ESLint rules to have a non-empty `meta.languages` array containing only strings.
@@ -33,3 +39,8 @@ module.exports = {
   },
 };
 ```
+
+## Further Reading
+
+- [ESLint v10.2.0 released](https://eslint.org/blog/2026/04/eslint-v10.2.0-released/)
+- [ESLint custom rule docs](https://eslint.org/docs/latest/extend/custom-rules)
--- a/docs/rules/require-meta-languages.md
+++ b/docs/rules/require-meta-languages.md
@@ -12,7 +12,7 @@ Dynamically computed values and unresolved spreads in `meta` are skipped, matchi
 
 ## Rule Details
 
-This rule requires ESLint rules to have a non-empty `meta.languages` array containing only strings.
+This rule requires ESLint rules to have a non-empty `meta.languages` array whose entries are valid language identifiers: either the `'*'` wildcard or a `'namespace/language'` form (such as `'js/js'`, `'json/json'`, or `'markdown/*'`), with no duplicates.
 
 Examples of **incorrect** code for this rule:
 
@@ -21,6 +21,20 @@ Examples of **incorrect** code for this rule:
 
 module.exports = {
   meta: {},
+  create(context) {
+    /* ... */
+  },
+};
+
+module.exports = {
+  meta: { languages: ['javascript'] },
+  create(context) {
+    /* ... */
+  },
+};
+
+module.exports = {
+  meta: { languages: ['js/js', 'js/js'] },
   create(context) {
     /* ... */
   },
@@ -34,6 +48,13 @@ Examples of **correct** code for this rule:
 
 module.exports = {
   meta: { languages: ['js/js'] },
+  create(context) {
+    /* ... */
+  },
+};
+
+module.exports = {
+  meta: { languages: ['markdown/*'] },
   create(context) {
     /* ... */
   },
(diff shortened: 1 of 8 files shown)


eslint.config.ts
```diff
@@ -66,6 +66,11 @@ export default defineConfig([
     plugins: { 'eslint-plugin': eslintPlugin },
     extends: ['eslint-plugin/all'],
     rules: {
+      // `require-meta-languages` targets `meta.languages` (added in ESLint v10.2.0).
```
bmish: Add a `TODO:` prefix.

docs/rules/require-meta-languages.md
bmish, line 1: > 
> 
> Thanks for taking this on — the implementation looks great. It mirrors `require-meta-type` closely (same `getRuleInfo` guard, `evaluateObjectProperties` / `hasUnresolvedObjectSpread` handling, and the ignore-non-static-values policy), which is exactly the right template. I also appreciated the manual testing against real-world plugins. A few requests before merging:
> 
> **1. Rule doc: please add motivation.**
> 
> Right now the doc jumps straight to examples. Since `meta.languages` is only a few months old, most plugin authors won't know what it is or why they should care. Could you add a short section explaining:
> 
> - What the property does: introduced in ESLint v10.2.0, it declares which languages a rule supports, using plugin-prefixed identifiers like `'js/js'` or `'json/json'`.
> - Why it matters: when omitted, ESLint assumes the rule supports *every* language. For almost all rules that's incorrect — a JS-assuming rule applied to a JSON or CSS file will crash on an unfamiliar AST or silently do nothing. Declaring `meta.languages` turns that into a clear config-time error instead.
> - That dynamically-computed values (and unresolved `meta` spreads) are skipped, matching the other `require-meta-*` rules.
> 
> A "Further Reading" section linking to the [v10.2.0 release post](https://eslint.org/blog/2026/04/eslint-v10.2.0-released/) and the custom-rules docs would round it out nicely.
> 
> **2. Tests: a few more shapes, for parity with the sibling rules.**
> 
> The suite covers the core cases, but `tests/lib/rules/require-meta-type.ts` exercises several shapes this one doesn't. Since the fixes are cheap, please add:
> 
> - ESM (`export default`) — one valid, one invalid
> - `meta` stored in a variable: `const meta = {}; module.exports = { meta, create(context) {} };`
> - A *resolved* spread supplying the property: `const extra = { languages: ['js/js'] }; module.exports = { meta: { ...extra }, create(context) {} };` (currently only the unresolved-spread skip is tested)
> - A static variable value: `const langs = ['js/js']; ... languages: langs` (valid), and an invalid variant confirming the report lands on the `Identifier`
> - `languages: null` and `languages: undefined` — these resolve statically, so they should report `invalid` rather than being skipped; worth locking that behavior in
> - Nit: the `'const x = 5;'` valid case is an unusual stand-in for "not a rule" — the sibling suites use `module.exports = {};` and a non-function `create`
> 
> **3. Non-blocking follow-up idea.**
> 
> The rule accepts any non-empty string array, so `languages: ['javascript']` (not a real registered identifier) passes. Validating the `pluginNamespace/languageName` shape and flagging duplicates could be a nice future enhancement — fine as a separate issue rather than this PR, especially since the property's semantics may still evolve.
> 
> Everything else — registration in `lib/index.ts`, README row, e2e fixture updates, and the `eslint.config.ts` self-exclusion with the explanatory comment — looks right. 👍
>

morgan-coded:
Pushed a revision addressing the review:

- **Doc:** added a motivation section — what `meta.languages` is (introduced in v10.2.0, plugin-prefixed identifiers like `'js/js'`), why omitting it is usually wrong (ESLint otherwise assumes every language, so a JS rule can crash or silently no-op on a JSON/CSS file), and a note that dynamic values and unresolved `meta` spreads are skipped like the other `require-meta-*` rules — plus a Further Reading section linking the v10.2.0 release post and the custom-rule docs.
- **Tests:** added the parity shapes from `require-meta-type` — ESM (valid + invalid), `meta` stored in a variable, a resolved spread supplying `languages`, a static array variable (valid) and a static string variable (invalid, reported on the `Identifier`), and `languages: null` / `undefined` (both reported as `invalid`). Also swapped the `const x = 5;` stand-in for `module.exports = {};` and a non-function `create`.
- **Config:** prefixed the `eslint.config.ts` self-exclusion comment with `TODO:`.

On the format/duplicate validation: makes sense as a follow-up — I'll open a separate issue for it so it can be discussed on its own, since the property's semantics may still evolve. Kept it out of this PR as you suggested.

tests/lib/rules/require-meta-languages.ts
bmish, line 1: >  the non-function-create valid test pairs it with a valid languages array, so it passes for two reasons at once — using meta: {} there (as the require-meta-type suite does) would prove it's the "not a rule" path being exercised.

lib/rules/require-meta-languages.ts
bmish, line 1: We could actually add some more validation now. Now is the best time since adding additional checks later would be breaking changes.

Ideas:

> 1. Reject empty-string entries. languages: [''] passes the current check (typeof '' === 'string'). This is arguably a latent bug in the PR as written, not even an enhancement.
> 2. Validate entry format. Each entry should be either the literal '*' (wildcard) or contain a / with non-empty segments (which covers both 'js/js' and 'markdown/*'). This catches the realistic mistake — ['javascript'] or ['js'] — which ESLint would otherwise turn into a confusing runtime mismatch. I'd deliberately stop at "has a slash with non-empty parts" rather than "exactly one slash": the docs only show short namespaces, and I couldn't confirm whether scoped forms like @eslint/json/json are legal, so the looser check avoids false positives on the unknown case.
> 3. Reject duplicate entries. ['js/js', 'js/js'] — zero false-positive risk, pure win.
> 4. Report the specific offending element, not the whole array, when one entry is bad. Not a behavioral expansion, but it's the kind of reporting-precision change that's easiest to do while the tests are being written anyway.

bmish (approved):
Thanks!

PR #648, opened 2026-07-22
feat: add new rules `no-incomplete-meta-schema` and `no-incorrect-meta-schema`

### LINKED ISSUE #296: New Rule: `no-incomplete-schema`
This could also potentially be part of [eslint-plugin/require-meta-schema](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-schema.md) but that might be too much in one rule.

The rule would check for common useless / no-op schemas as discussed in https://github.com/eslint/rfcs/pull/85#discussion_r775144472 such as:
* `{}`
* `{ type: "array" }`
* etc

And also catch stuff like:
* Expecting an object with `{ type: 'object', properties: { ... } }` without using `additionalProperties: false` to ensure all properties are included

morgan-coded:
Fixes #296.

This splits rule-option schema analysis into an objective correctness rule and a separate opt-in policy rule.

## Correctness rule

`no-incorrect-meta-schema` enables ten checks by default when the rule is enabled: `emptyRoot`, `bareArrayRoot`, `nonArrayRootType`, `nonConstrainingRoot`, `ignoredKeywords`, `ignoredRefSiblings`, `unresolvedRefs`, `ignoredAdditionalItems`, `incompatibleTypeKeywords`, and `impossibleBounds`.

The ignored-keyword check uses a curated inert set rather than rejecting unknown keywords. It covers later-draft keywords, the confusable JTD/OpenAPI entries `elements`, `values`, `optionalProperties`, and `discriminator`, draft-03 `extends`, `disallow`, and `divisibleBy`, and degenerate `allOf: []`. It traverses live conditional and negative schemas and statically resolvable schema containers.

Reference checks model ESLint's array-form rewrite as a wrapper object around positional schemas. Fragment pointers resolve from the active resource root; statically declared `id` and `$id` values index named fragments and embedded resources, including pointers into those resources, while genuinely external references remain unresolved under Ajv's `missingRefs: "ignore"` behavior.

## Policy rule

`no-incomplete-meta-schema` remains opt-in and keeps four checks: `explicitAdditionalProperties`, `explicitItems`, `typedItems`, and `boundedTuples`. It does not descend into negative or conditional branches, and bare `if` or `not` schemas do not count as an item policy.

Both rules expose named `checks` options and remain outside `recommended` in this minor release. The correctness rule is the safe candidate for a future `recommended` config; the policy rule is intended to remain opt-in.

## Dialect and corpus

The behavior follows ESLint's Ajv 6 setup: draft-04 meta-schema, draft-07 runtime keywords, and `schemaId: "auto"`.

The post-fix recount is:

| Corpus | `no-incorrect-meta-schema` | `no-incomplete-meta-schema` |
| --- | ---: | ---: |
| ESLint core (292) | 1 | 6 |
| this plugin (40 rule files) | 0 | 0 |

The core correctness finding is `no-restricted-modules` using ignored `additionalItems`; the policy findings remain opt-in authoring guidance. The plugin's own rule files stay clean with both rules enabled.


morgan-coded:
CI note: all 12 failing jobs stop in `npm install` with npm's internal `Cannot read properties of null (reading 'edgesOut')` before any test runs, on both the original run and a retrigger. The branch doesn't touch `package.json` or the lockfile, and the same install plus the full suite pass locally on these manifests. Main's last scheduled run (July 21) was green, so this looks like the runners' npm hitting something that changed in the registry window since — not PR-attributable as far as I can tell, but happy to adjust if it turns out to be branch-side.

lib/rules/no-incomplete-schema.ts
bmish, line 1: AI feedback from Grok/Sol:

### 1. Disabling a root check also skips nested checks

At `lib/rules/no-incomplete-schema.ts:659-664`, the rule returns after identifying a root defect even when that check is disabled.

For example, with `rootWrongOptionsType: false`, nested `missingItems` diagnostics under an object-typed root disappear. Traversal should continue when the root check is disabled.

### 2. `prefixItems` is treated as effective, but ESLint ignores it

`hasItemConstraint` accepts `prefixItems`, and the tests consider this complete:

```js
schema: [{ type: 'array', prefixItems: [{ type: 'string' }] }]
```

Current ESLint option validation accepts both string and numeric first items here because it ignores `prefixItems`. The rule should follow ESLint’s actual schema dialect rather than treating newer-draft keywords as effective constraints.

### 3. Boolean and schema-valued item handling is incorrect

`items: true` reports nothing despite accepting every item:

```js
schema: [{ type: 'array', items: true }]
```

Conversely, schema-valued `additionalItems` reports `unboundedTuple`:

```js
schema: [{
  type: 'array',
  items: [{ type: 'string' }],
  additionalItems: { type: 'number' },
}]
```

Those extra items are constrained. If the intended policy is strictly fixed-length tuples, the documentation should say that; as currently documented, this is a false positive. `items: false` should remain valid.

### 4. Recursive checks create false positives in conditional and negative schemas

For example:

```js
schema: [{ not: { type: 'array' } }]
```

This correctly rejects arrays, but the rule reports `missingItems` on the inner schema. Adding typed `items` would change its meaning.

The same issue applies to `if`/`then`/`else` and object checks such as `additionalProperties`. These subschemas cannot always be evaluated as standalone positive option constraints.

### 5. Array-form traversal has gaps

These no-op positional schemas are silent:

```js
schema: [{}]
schema: [{ title: 'description only' }]
```

Statically resolvable local schemas are also skipped:

```js
const option = {
  type: 'object',
  properties: { value: { type: 'string' } },
};

schema: [option]
```

`isStaticallyInspectable` accepts this, but traversal only visits direct `ObjectExpression` elements.


## Minor

- Reword the excluded policy checks so they are clearly “considered but not shipped,” rather than appearing to be configurable options.
- `type: ['array']` produces `missingItems`, while `type: 'array'` produces `bareArraySchema`; aligning those would improve consistency.
- Add regression coverage for disabled root checks, conditional/negative subschemas, boolean schemas, and multiple diagnostics.

docs/rules/no-incomplete-schema.md
bmish, line 1: Discussion about what checks to include by default.

morgan-coded:
Addressed all seven items in fb6398ca — details below, one pushback included.

Verifying item 2 led to `eslint/lib/shared/ajv.js`: ESLint 9 constructs ajv 6 with the draft-04 meta-schema, so the enforceable dialect is ajv 6's draft-07 runtime set. The rule now treats `prefixItems`, `unevaluatedItems`/`unevaluatedProperties`, `$dynamicRef`/`$recursiveRef`, `minContains`/`maxContains`, `dependentSchemas`/`dependentRequired`, and `contentSchema` as ignored keywords — they no longer credit a schema as constrained and traversal no longer descends into them — and the doc gained a dialect-basis note.

Item 1: a disabled root check now falls through to nested traversal (regression test drives `rootWrongOptionsType: false`). Item 3: `items: true` reports `missingItems`, `items: false` stays valid, and schema-valued `additionalItems` no longer reports `unboundedTuple`. Item 4: `not`/`if`/`then`/`else` are out of positive-check traversal; I kept `additionalProperties` descent because ajv evaluates that subschema against every unmatched member, so constraining it narrows what the option accepts — a test locks the behavior, and I'll drop it too if you read it differently. Item 5: statically resolvable positional elements now resolve through the same machinery as the root, and no-op elements report (`{}` as `emptySchema`, annotation-only as `ineffectiveRootSchema`). Minors: excluded policies reworded as considered-but-not-shipped, `type: ['array']` aligned to `bareArraySchema`, and regression coverage added for each class plus a four-diagnostic case.

On which checks to include by default: everything currently ships enabled, per the reasonably-complete-while-safe framing in #296 — each default-on check fires only where this ajv provably ignores the schema or accepts it unbounded, and intentional validation-skip forms remain expressible via `false` or per-check disables. If you'd rather launch with a trimmed default set, name the cut and I'll flip those to opt-in.

Local gates are green: 64 focused and 1336 full tests, lint, typecheck, format, generated-docs check, and both e2e suites; the CI install crash noted earlier is unchanged and unrelated.

docs/rules/no-incomplete-schema.md
bmish: I did some research with AI about what checks to include. I'm thinking we can actually split the rule in two. One rule focused on safe checks that we can make `recommended`, and then another rule focused on more opinionated checks that would be opt-in.

---

I ran my own pass over this, mostly to pressure-test the default-set question, and compared
the result with [Ajv strict mode](https://ajv.js.org/strict-mode.html) and the community's
[JSON Schema lint-rule catalog](https://github.com/Karan-Palan/json-schema-lint-rules).
Their correctness-vs-policy split looks like the right model here, and I think it is strong
enough to be the rule boundary:

- **`no-useless-meta-schema`**: objective defects where a schema or keyword is ignored,
  accepts everything, or makes the options schema unusable. This is a candidate for the
  recommended config in the next major release.
- **`no-incomplete-meta-schema`**: valid but potentially underspecified schemas. This remains
  opt-in.

Both rules should retain a named `checks` option. That gives users individual control and
lets us add checks later default-off in a minor release, then consider enabling them by
default in a major release.

For consistency, checks in the useless rule should name the defect (`ignoredKeywords`,
`unresolvedRefs`), while checks in the incomplete rule should name the desired property
(`explicitItems`, `boundedTuples`).

## Recommendation / merge bar

### `no-useless-meta-schema`: correctness

- `emptyRoot`, `bareArrayRoot`, `nonArrayRootType`, and `nonConstrainingRoot`, after making
  the effective-keyword list match ESLint
- `ignoredKeywords` — a curated list of constructs ESLint silently ignores,
  including `prefixItems`, `unevaluatedItems`, `unevaluatedProperties`,
  `dependentRequired`, `dependentSchemas`, `minContains`, `maxContains`, `$dynamicRef`,
  `$recursiveRef`, `contentSchema`, and draft-03 `required: true`
- `ignoredRefSiblings`
- `unresolvedRefs`
- `ignoredAdditionalItems` (`items` absent or not tuple-form)
- `incompatibleTypeKeywords`
- `impossibleBounds` (`minItems > maxItems`, etc.)

These checks should default on when the rule is enabled. Keep the rule unrecommended in the
current minor release, with the goal of adding it to `recommended` in the next major.

### `no-incomplete-meta-schema`: completeness policy

- `explicitAdditionalProperties`
- `explicitItems`, `typedItems`
- `boundedTuples`
- mandatory `type`, `required`, `minItems`/`maxItems`, `uniqueItems`, string bounds,
  `additionalProperties: false`, and blanket unknown-keyword rejection

This rule remains unrecommended. Its initial checks can default on once a user opts into the
rule, while the `checks` option allows selectively disabling them. Newly added checks should
start default-off to avoid adding diagnostics for existing users in a minor release.

### Implementation fixes before merge

1. Split the current implementation and documentation along the useless-vs-incomplete
   boundary above.
2. When a root check is disabled, do not return before visiting nested schemas. Today the
   root-defect branch suppresses all nested checks regardless of whether it reports.
3. Do not count keywords that ESLint ignores as satisfying constraints. In particular,
   `prefixItems` currently makes `explicitItems` pass and has a valid test asserting that
   result.
4. For a full object-form root, remember that ESLint always validates `context.options`, which
   is an array. Root array checks should not require an explicit `type: 'array'`.

The rest of this comment is the supporting evidence for that split.

## Why this split matches ESLint's validator

The precedent points in the same direction. [ESLint RFC 85](https://github.com/eslint/rfcs/blob/main/designs/2021-schema-object-rules/README.md#no-op-schemas)
distinguishes explicit opt-out (`schema: false`) from fake/no-op schemas. Ajv strict mode
treats ignored keywords as correctness problems while presenting strict types and tuples as
optional restrictions. The community catalog similarly categorizes ignored
`additionalItems` and incompatible type keywords as correctness, but explicit
`additionalProperties` and collection bounds as style/opinionated.

Current ESLint uses Ajv 6 with a draft-04 metaschema, `validateSchema: false`, and
`missingRefs: "ignore"`. In practice, that means schemas are not meta-schema validated,
missing references accept everything, draft-07 keywords work, and later-draft keywords are
silently ignored.

I tested schemas against ESLint's configured Ajv rather than classifying keywords by name:

- `items`, `const`, `contains`, `if`/`then`, `propertyNames`, and local `$ref`s constrain.
- `prefixItems`, `unevaluatedItems`, `unevaluatedProperties`, `dependentRequired`,
  `dependentSchemas`, `minContains`, `maxContains`, and draft-03 `required: true` do not.
- `additionalItems` is ignored unless `items` is tuple-form.
- `$ref` siblings and unresolved references are ignored.

That gives the correctness rule an objective test: *does this construct constrain rule options
under ESLint's validator?* It also explains why a curated ineffective-keyword list is safer
than blanket unknown-keyword rejection, since extension annotations are legal.

This is observable in the ecosystem: two `eslint-plugin-n` rules use `prefixItems` plus an
ineffective `additionalItems`, allowing values the schema appears to reject; the current rule
reports those schemas clean. Across 292 core and 401 third-party schemas, the proposed
correctness checks found 12 ineffective `additionalItems` uses and two ignored-keyword uses,
with no false positives in the sweep. `$ref` sibling, unresolved-ref, type-incompatibility,
and impossible-bound checks had no hits.

## The `type: 'array'` gate inverts the incentive

`boundedTuples`, `explicitItems`, and `typedItems` all gate on
`hasType(properties, 'array')`. For an object-form root that's the wrong gate — we already
know the value being validated is the options array, so no explicit `type` is needed. Adding
`type: 'array'` turns the checks on while omitting it turns them off. Core's `func-names`
schema is treated as valid despite accepting a bogus third option, and 15 of 24 object-form
core schemas omit the root type. Drop this gate at the object-form root and keep it for nested
subschemas.

Do not extend `emptyRoot` to `schema: [{}]`: ESLint's array shorthand adds cardinality
semantics, so that schema accepts at most one arbitrary option and is not equivalent to root
`schema: {}`.

## What belongs in `no-incomplete-meta-schema`

The four non-root checks are useful completeness policies, but they are not correctness
checks. They belong in the opt-in rule:

- **`typedItems` + `explicitItems` contradict each other.** `explicitItems` demands `items`;
  the natural "array of anything" spelling `items: {}` then trips `typedItems`. There is
  no way to express an intentionally unconstrained array with both on. If both remain
  available, `items: {}` should be treated as an explicit opt-in and each check should be
  individually disableable.
- **`boundedTuples` has no opt-in spelling.** `additionalItems: true` still reports,
  so an intentionally open tuple can't be declared. Accept explicit `additionalItems: true`
  the way `explicitAdditionalProperties` accepts `additionalProperties: true`, or
  leave this particular check default-off within the opt-in rule.

`explicitAdditionalProperties` has a behaviour-preserving fix because
`true`/`false`/schema all satisfy it, but requiring explicitness is still an authoring policy,
so it belongs in the incomplete rule. Keep it local to the subschema: `additionalProperties`
only accounts for sibling `properties` and `patternProperties`, so treating a declaration
inside `allOf` or a referenced schema as closing the outer object can be wrong.

## Smaller items

1. **Corpus claim.** The PR says the launch set reports zero findings on the 292 core rule
   schemas. I get **5 findings in 2 files** — `missingAdditionalProperties` ×4 in
   `indent-legacy`, `unboundedTuple` ×1 in `array-element-newline`. Both look like true
   positives (`array-element-newline` really does accept
   `['always','always',{totally:'bogus'}]`), so this is good news for the rule — but the
   default-set argument leans on that sweep, so it's worth reconciling how we got different
   numbers.

2. **`typedItems`' unknown-keyword bail-out is a soundness hole.** It skips reporting when
   the item schema has any unrecognized key, so `items: { itemz: { type: 'string' } }`
   comes back clean.

3. **Docs.** The closing paragraph listing rejected policy checks reads as design notes; I'd
   move it to the PR description or the issue.
   Also worth documenting *why* a check exists — a one-line "this keyword does nothing under
   ESLint's configured Ajv" per check is the most useful thing for someone hitting a report.

Nice work on this overall — the root no-op checks are solid, and the `schema: false` handling
is the right call. Splitting objective defects into `no-useless-meta-schema` and optional
completeness policy into `no-incomplete-meta-schema` gives the recommended config a safe
future path without losing per-check control or room to add checks later.

morgan-coded:
Split landed in 1f93354 along your correctness-versus-policy boundary: `no-useless-meta-schema` carries the four root checks plus `ignoredKeywords` (including draft-03 `required: true`), `ignoredRefSiblings`, `unresolvedRefs`, `ignoredAdditionalItems`, `incompatibleTypeKeywords`, and `impossibleBounds`, all default-on; `no-incomplete-meta-schema` keeps `explicitAdditionalProperties`, `explicitItems`, `typedItems`, and `boundedTuples`. Both stay unrecommended in this minor and both retain the named `checks` option, defect-named on the useless side and desired-property-named on the incomplete side.

Your implementation fixes 2 and 3 were already in fb6398ca, which landed a few minutes before your review posted. This round adds fix 4 — object-form roots are checked without requiring `type: 'array'`, with your `func-names` shape as a fixture — restores `[{}]` and annotation-only positional schemas as non-useless per the cardinality point, treats `items: {}` and `additionalItems: true` as explicit opt-ins so the check pairs can't contradict, and closes the `typedItems` unknown-keyword bailout.

On the corpus numbers: rerunning all 292 core schemas with the split rules reproduces your five findings exactly, plus the expected `func-names` root-gate hit and an ignored-`additionalItems` hit in `no-restricted-modules` — 7 findings across 4 files. The original zero claim was a reporting error on my side; the sweep output already contained your five.

I read the mandatory `type`/`required`/bounds/`additionalProperties: false`/blanket-unknown-keyword list as the future default-off ladder under the minor/major model rather than merge scope — say the word if you meant any of them now. Local gates: 1,391 tests, coverage 98.36/96.66/99.27/98.46 against the 95/93/95/95 thresholds, lint, typecheck, format, generated docs, and both e2e suites.

Design notes moved here from the rule doc per your round-1 note: the policy checks considered during initial design but not shipped were `schemaType`, `knownKeywords`, `objectAdditionalPropertiesFalse`, `arrayMinItems`/`arrayMaxItems`, `stringMinLength`/`stringMaxLength`, `objectProperties`, and `arrayUniqueItems` — dropped because valid schemas produced false positives for those policies; under the split they are the natural candidates for later default-off additions to the incomplete rule.

docs/rules/no-incomplete-meta-schema.md
```diff
@@ -0,0 +1,86 @@
...
+  },
+};
+```
+
+`items: {}` is an explicit opt-in to unconstrained items.
+`additionalItems: true` is an explicit opt-in to an open tuple.
+
+## Options
+
+All four checks are enabled when this opt-in rule is enabled. Each check can
```
bmish: "All four checks are enabled when this opt-in rule is enabled." we can remove this line as it's just going to become outdated when more checks are added in the future.

docs/rules/no-incomplete-meta-schema.md
bmish, line 1: We can add a `Related Rules` section to each rule doc to link to the other rule. We should also make clear that one of the rules is intended to be fully safe and become a `recommended` rule in the future, while the other is intended to be more opinionated and will not become `recommended`.

eslint.config.ts
```diff
@@ -81,6 +81,21 @@ export default defineConfig([
       ],
     },
   },
+  {
+    // TODO: drop this block once the follow-up PR for #649 switches `elements` -> `items`
```
bmish: 649 has merged.

docs/rules/no-incomplete-meta-schema.md
bmish, line 1: Another round of feedback:

----


The split landed well. I re-ran my harness against `610ae28` and confirmed the previous round is genuinely addressed — `prefixItems` and the real `eslint-plugin-n` shape now report `ignoredKeywords` + `ignoredAdditionalItems`, the object-form root `type` gate is gone (`func-names` now reports `boundedTuples`), `items: {}` and `additionalItems: true` work as opt-in escapes, and `impossibleBounds` / `ignoredRefSiblings` / draft-03 `required: true` are all caught. Corpus numbers on my side:

| Corpus            | `no-useless-meta-schema`                             | `no-incomplete-meta-schema` |
| ----------------- | ---------------------------------------------------- | --------------------------- |
| ESLint core (292) | 1 (`no-restricted-modules` `ignoredAdditionalItems`) | 9                           |
| this plugin (40)  | 0                                                    | 0                           |

That's the profile we wanted from the correctness rule. 1393 tests pass locally at `610ae28`.

Three correctness gaps before merge, all of which I reproduced.

## 1. Blocking — `$ref` resolution doesn't model ESLint's array-form rewrite

ESLint rewrites `schema: [...]` into `{ type: 'array', items: [...], minItems: 0, maxItems: n }` (`eslint/lib/config/config.js:198-205`), so the resolution root is the synthesized wrapper, not each positional schema. The rule uses the element as its own root (`no-useless-meta-schema.ts:599-605`), which is wrong in both directions. Measured against ESLint's actual Ajv:

| Array-form schema                                                                         | Ajv                                                | rule                |
| ----------------------------------------------------------------------------------------- | -------------------------------------------------- | ------------------- |
| `[{definitions:{v:…}, $ref:'#/definitions/v'}]`                                           | **accepts everything** (`can't resolve reference`) | clean ❌            |
| `[{definitions:{v:…}, type:'array', items:{$ref:'#/items/0/definitions/v'}}]`             | resolves                                           | `unresolvedRefs` ❌ |
| `[{id:'http://ex/s.json', definitions:{v:…}, …, $ref:'http://ex/s.json#/definitions/v'}]` | resolves                                           | `unresolvedRefs` ❌ |

Two things I'd add to the fix beyond re-rooting:

- **Base-URI awareness is part of it.** Absolute self-references resolve in Ajv whenever a root `id`/`$id` is present — in object-form *and* array-form (the `id` in an element establishes the base URI). `resolveLocalReference` rejects every non-fragment URI, so this needs an `id`/`$id` check, not just a different root object. Genuinely external refs stay reportable — those really are no-ops under `missingRefs: "ignore"`.
- **`#/items/0/...` is not universally a valid escape.** When the `$ref` *is* the element root, Ajv self-recurses and `compile` throws `Maximum call stack size exceeded` — that spelling crashes ESLint at config-validation time rather than no-op'ing. It resolves fine when the `$ref` sits nested under the element or in a sibling element. Probably worth its own message eventually; at minimum don't recommend `#/items/0/...` in docs as the fix.

Sizing note, since it affects how hard to push on this: I swept every `$ref`-using schema in core plus unicorn/n/es-x/eslint-comments — **6 rules, all object-form, all resolving**. So today this is latent in both directions rather than live breakage. Still worth fixing before the rule heads for `recommended`, since the wrong model is the thing that's hard to change later, but I wouldn't hold the PR on the frequency argument alone.

## 2. Blocking — correctness traversal skips `not` / `if` / `then` / `else`

`getChildSchemas` (`meta-schema-utils.ts:181-245`) handles `additionalItems`, `additionalProperties`, `contains`, `propertyNames`, `items`, `extends`, `allOf`/`anyOf`/`oneOf`, `$defs`/`definitions`/`patternProperties`/`properties`, and `dependencies` — but not `not`/`if`/`then`/`else`. These *were* in `directSchemaKeywords` before the split, so this is a regression the split introduced rather than a pre-existing gap.

All four are live keywords under Ajv 6 (I verified `if`/`then` and `not` both constrain), so anything nested inside them is unreachable by the correctness checks. Confirmed clean, all four:

```js
schema: [{ not: { prefixItems: [{ type: 'string' }] } }]
schema: [{ type: 'object', if: { prefixItems: […] }, then: { required: ['a'] } }]
schema: [{ type: 'object', if: { required: ['a'] }, then: { unevaluatedProperties: false } }]
schema: [{ type: 'object', if: { required: ['a'] }, else: { $ref: '#/nope' } }]
```

The first is the sharp one: Ajv drops `prefixItems`, leaving `not: {}`, which rejects every value. Precisely — `[]` still passes (the wrapper's `minItems: 0` means item 0 is never reached), so the rule can be *enabled* but can never be *configured*. Worth stating that way in the eventual test case.

Agreed the two rules need different child-traversal policies — skipping these for completeness policy is right, skipping them for correctness isn't.

## 3. Correctness gap — statically resolvable containers are skipped

`isStaticallyInspectable` accepts them but `getChildSchemas` only walks direct array/object expressions, so hoisting a container into a constant silently disables the checks:

```js
const properties = { value: { prefixItems: [] } };
schema: [{ type: 'object', properties }]                  // clean

const items = [{ prefixItems: [] }];
schema: [{ type: 'array', items }]                        // clean

const allOf = [{ prefixItems: [] }];
schema: [{ allOf }]                                       // clean

schema: [{ type: 'object', properties: { value: { prefixItems: [] } } }]  // ignoredKeywords ✅
```

`getChildSchemas` already calls `resolveObjectExpression` for the leaves — the container lookups need the same treatment.

## 4. New — `elements` isn't in the curated ignored-keyword list

`ignoredByAjv6Keywords` (`meta-schema-utils.ts:34-45`) covers only draft-2019/2020 keywords. JTD keywords aren't in it, so the exact bug that motivated this PR isn't caught by the correctness rule:

```js
// the no-property-in-node shape
schema: [{ type: 'object', properties: { f: { type: 'array', elements: { type: 'string' } } }, additionalProperties: false }]
//   no-useless-meta-schema     -> clean          ← the motivating bug
//   no-incomplete-meta-schema  -> explicitItems  ← only via the opt-in rule
```

`values` and `optionalProperties` are likewise clean under both rules. Since `no-useless-meta-schema` is the one aimed at `recommended`, I'd add `elements`, `values`, `optionalProperties`, and `discriminator` to the curated list. That stays targeted — it's a short list of known-confusable keywords, not the blanket unknown-keyword rejection we already ruled out.

## Smaller items

- **Non-blocking, agreed:** `hasItemConstraint` treats bare `if` and `not` as item policies. Confirmed — `{type:'array', if:{minItems:1}}` and `{type:'array', not:{maxItems:0}}` both suppress `explicitItems`, though neither says anything about items.

- **The `eslint.config.ts` carve-out is now provably dead.** The branch merged `main`, so #650's `elements` → `items` fix is in and both `order` rules are clean — but the `files: ['lib/rules/meta-property-ordering.ts', 'lib/rules/test-case-property-ordering.ts']` block survived the merge, still disabling `explicitItems`, with a TODO describing a change that already happened. Self-linting this plugin reports zero findings with the block removed, so it can just go.

- **PR body is stale** (title looks updated already): it still describes a single `no-incomplete-schema` rule with an `arrayItems` check, still references the obsolete #649 carve-out, and still carries the zero-findings-on-both-corpora claim. Feel free to use the table at the top for the corrected numbers.

- **CI:** agreed the `edgesOut` failures look like npm rather than the branch — the merge from `main` didn't shift them, and the suite is green locally.

Overall this is close. #2 and #3 look mechanical; #1 is the one worth getting exactly right since the resolution model is what we'd be stuck with.

bmish, line 1: I had my AI perform a real-world sanity check on popular eslint plugins of these rules. Results seem good.

---

## Real-world sanity check

I ran both new rules (defaults only) against the 16 plugins listed under [eslint-doc-generator § Users](https://github.com/eslint-community/eslint-doc-generator#users), focused on rule implementation files.

**Result: looks good — no changes needed from this sweep.**

| Rule | Findings | Notes |
| --- | ---: | --- |
| `no-useless-meta-schema` | 5 | All `ignoredAdditionalItems` — real no-ops (`additionalItems: false` beside non-tuple `items`) in `eslint-plugin-n` and `eslint-plugin-jest` |
| `no-incomplete-meta-schema` | 27 | Mostly bare `type: 'array'` without `items` (unicorn alone accounts for 17); fair opt-in policy hits, not FPs |

9 of 16 plugins were clean (`ember`, `import`, `jest-dom`, `jsx-a11y`, `promise`, `qunit`, `react`, `regexp`, `security`). No crashes, and no noisy hits from the heavier useless checks (`emptyRoot`, `ignoredKeywords`, `unresolvedRefs`, etc.).

That matches the intended split: useless findings are sparse/objective; incomplete findings are common authoring gaps and stay appropriately opt-in. I wouldn’t adjust defaults based on this.

bmish: One more set of changes for v1, all in §4's territory.

**Extend the inert-keyword list — and fix three misclassifications.** Also inert under ESLint's Ajv (measured by comparing accept-sets with and without each keyword): draft-03 `extends`, `disallow`, `divisibleBy`, plus degenerate `allOf: []`.

The draft-03 three aren't merely missing from `ignoredByAjv6Keywords` — they're in `knownKeywords`, so they're whitelisted as *effective* and skip `nonConstrainingRoot`'s unknown-keyword path as well. `getChildSchemas` also traverses `extends` as a live composition keyword. Worth making the partition explicit — every keyword in exactly one of effective / inert / unknown — so the three sets can't drift apart again.

**One trap when implementing:** don't treat keys inside `properties` / `patternProperties` / `definitions` as keywords. `keyword-spacing` has an option literally named `extends`, and a naive walker reports it as a dead draft-03 keyword. The current `getChildSchemas` recurses into map *values*, so it's already right — just don't lose that when adding the check.

Priority: **zero hits across 693 real schemas** (core 292 + unicorn 141 + n 43 + es-x 208 + eslint-comments 9) for all of the above, so these are cheap-to-add rather than urgent. `elements` stays the only one with demonstrated incidence.

**No change wanted on the policy rule's scope** — four checks is right for v1. The #296 strict items (`minItems: 1`, `uniqueItems`, string bounds, mandatory `type`) can land default-off later via `checks` without a breaking change.

Two v2 candidates I'm explicitly *not* asking for now, just so they're on record: unsatisfiable schemas beyond `impossibleBounds` (`enum: []`, `not: {}`, `type`/`const` conflicts, disjoint `allOf`), and unknown `format` — which throws at compile rather than no-op'ing, making it the only failure mode here that crashes instead of silently passing.

morgan-coded:
Reference resolution now models ESLint's synthesized array wrapper: fragment pointers resolve from the active wrapper/resource root, `id`/`$id` establish base URIs so matching absolute self-references resolve, genuinely external documents still report, and the self-recursive element-root spelling stays safe and silent rather than crashing traversal. The correctness rule traverses `not`/`if`/`then`/`else` again and resolves hoisted schema containers; the policy rule keeps its four-check scope and still skips those branches. The keyword model is one explicit effective/inert/unknown partition now — a single typed source generates both sets, a startup invariant rejects double-listing, and a regression test pins the boundary — with the draft-03 trio moved out of effective, `extends` no longer traversed, empty `allOf` reported, the JTD/OpenAPI four inert, and map option names like keyword-spacing's `extends` untouched. Bare `if`/`not` no longer suppress `explicitItems`.

This push also carries the earlier docs round: the check-count line is gone, both docs cross-link via Related Rules with the future-`recommended` versus permanently-opt-in framing, and the dead ordering-rules carve-out is dropped — self-lint passes with no exclusions.

The post-fix recount matches your sweep exactly (core 1/9, this plugin 0/0), and the PR description is refreshed to describe the split, the dialect basis, the wrapper reference model, and your corpus table. Local gates: 1,424 tests, coverage 97.74/95.98/99.29/97.84, lint, typecheck, format, generated docs, both e2e suites. Your two v2 candidates are noted as on-record for later.

lib/rules/no-incorrect-meta-schema.ts
```diff
@@ -0,0 +1,666 @@
...
+          );
+
+          if (checks.ignoredKeywords && hasIgnoredKeyword(properties, scope)) {
+            context.report({ node: schema, messageId: 'ignoredKeywords' });
+          }
+          if (checks.ignoredRefSiblings && hasIgnoredRefSibling(properties)) {
+            context.report({ node: schema, messageId: 'ignoredRefSiblings' });
+          }
+          if (
+            checks.unresolvedRefs &&
```
bmish, line 746: All previously raised items are addressed: wrapper refs, conditional traversal, hoisted containers, bare `if`/`not`, PR metadata, and docs.

One new blocker remains: `unresolvedRefs` rejects valid Draft-04 references using named `id` fragments or embedded resources.

```js
schema: {
  id: 'http://example.com/root.json',
  type: 'array',
  definitions: {
    value: { id: '#value', type: 'string' },
  },
  items: { $ref: '#value' },
}
```

ESLint’s Ajv resolves this and enforces the string type; the rule reports `unresolvedRefs`. The same false positive occurs when referencing a nested schema with `id: 'item.json'`. The resolver currently supports JSON Pointer fragments and self-document URIs, but does not index named IDs/resources (`no-useless-meta-schema.ts:177-215`).

Smaller, non-blocking: hoisted `allOf` branches still produce an `explicitItems` false positive in the policy rule:

```js
const branch = { items: {} };
schema: [{ type: 'array', allOf: [branch] }]
```

docs/rules/no-incomplete-meta-schema.md
```diff
@@ -0,0 +1,92 @@
+# eslint-plugin/no-incomplete-meta-schema
+
+📝 Require explicit policy choices in rule options schemas.
+
+<!-- end auto-generated rule header -->
+
+Rule option schemas can be valid while leaving important authoring policies
+implicit. This opt-in rule reports four places where a schema author may want
```
bmish: Let's not mention the number here as that will become outdated when additional checks are added later.

```suggestion
implicit. This opt-in rule reports places where a schema author may want
```

docs/rules/no-incomplete-meta-schema.md
```diff
@@ -0,0 +1,92 @@
...
+        { checks: { explicitAdditionalProperties: false } },
+      ],
+    },
+  },
+];
+```
+
+## Related Rules
+
+- [no-useless-meta-schema](./no-useless-meta-schema.md) — the safe counterpart. It reports only objective defects (schemas or keywords that ESLint's validator ignores or that accept everything) and is intended to become part of the `recommended` config in a future major release. This rule is deliberately opinionated completeness policy and will remain opt-in.
```
bmish: Let's avoid trying to mention too much about the other rule here to avoid repeating the other rule doc.

```suggestion
- [no-useless-meta-schema](./no-useless-meta-schema.md) — the safe counterpart. It reports only objective defects.
```

And let's move "This rule is deliberately opinionated completeness policy and will remain opt-in." higher int his rule doc, not as a side mention in related rules.

docs/rules/no-useless-meta-schema.md
```diff
@@ -0,0 +1,71 @@
...
+        { checks: { ignoredKeywords: false } },
+      ],
+    },
+  },
+];
+```
+
+## Related Rules
+
+- [no-incomplete-meta-schema](./no-incomplete-meta-schema.md) — the opinionated counterpart. It enforces opt-in completeness policies for schemas that are valid but underspecified, and it will remain opt-in. This rule is limited to objective defects and is intended to become part of the `recommended` config in a future major release.
```
bmish:
```suggestion
- [no-incomplete-meta-schema](./no-incomplete-meta-schema.md) — the opinionated counterpart.
```

docs/rules/no-useless-meta-schema.md
```diff
@@ -0,0 +1,71 @@
...
+
+`schema: false` remains the explicit validation opt-out established by
+[ESLint RFC 85](https://github.com/eslint/rfcs/tree/main/designs/2021-schema-object-rules#opt-out).
+In contrast, `schema: []` accepts no options, while `schema: [{}]` carries
+array-shorthand cardinality semantics and is not classified as useless.
+
+## Options
+
+All checks are enabled by default when this rule is enabled. The rule remains
+unrecommended in the current minor release. Each check can be disabled
```
bmish: No need to mention "The rule remains unrecommended in the current minor release" as that's going to become out-of-date eventually. But we should add a sentence higher up about this rule being intended to become part of the `recommended` config.

docs/rules/no-incomplete-meta-schema.md
```diff
@@ -0,0 +1,92 @@
...
+    /* ... */
+  },
+};
+```
+
+`items: {}` is an explicit opt-in to unconstrained items.
+`additionalItems: true` is an explicit opt-in to an open tuple.
+Bare `if` and `not` schemas do not state an item policy by themselves.
+Conditional and negative child schemas remain outside this policy rule's
+positive-schema traversal.
```
bmish: This seems too obscure, suggested rewording:

```suggestion
### Intentional open policies

This rule requires policies to be explicit; it does not require them to be
restrictive:

- Use `items: {}` to explicitly allow any array item.
- For tuple schemas, use `additionalItems: true` to explicitly allow items
  beyond the declared tuple positions.

Completeness checks are not applied inside `if`, `then`, `else`, or `not`,
because adding constraints there could change the schema's meaning.
```

docs/rules/no-incomplete-meta-schema.md
```diff
@@ -0,0 +1,92 @@
+# eslint-plugin/no-incomplete-meta-schema
+
+📝 Require explicit policy choices in rule options schemas.
+
+<!-- end auto-generated rule header -->
+
+Rule option schemas can be valid while leaving important authoring policies
+implicit. This opt-in rule reports four places where a schema author may want
+to state that policy explicitly.
```
bmish:
```suggestion
Rule option schemas can be valid while leaving important authoring choices
implicit. This opinionated rule is intended to remain opt-in and reports places
where schema authors may want to state those choices explicitly.
```

docs/rules/no-useless-meta-schema.md
```diff
@@ -0,0 +1,71 @@
...
+For `$ref` checks, object-form schemas use their schema object as the resolution
+root. Array-form schemas use the wrapper object that ESLint synthesizes around
+the positional schema array. An `id` or `$id` establishes a resource base for
+matching absolute self-references; references to other documents remain
+unresolved under ESLint's `missingRefs: 'ignore'` configuration.
+
+`schema: false` remains the explicit validation opt-out established by
+[ESLint RFC 85](https://github.com/eslint/rfcs/tree/main/designs/2021-schema-object-rules#opt-out).
+In contrast, `schema: []` accepts no options, while `schema: [{}]` carries
+array-shorthand cardinality semantics and is not classified as useless.
```
bmish: This section seems too obscure. Suggested replacement:

```suggestion
### Validator compatibility

This rule follows the schema keywords that ESLint actually enforces. ESLint
uses Ajv 6, which silently ignores some keywords from newer JSON Schema drafts
or other schema dialects. Examples include `prefixItems`,
`unevaluatedProperties`, `elements`, and `discriminator`.

The `ignoredKeywords` check reports only known ineffective constructs; it does
not reject arbitrary unknown keywords. It also reports known no-op forms such
as `allOf: []`.
```

docs/rules/no-incorrect-meta-schema.md
```diff
@@ -0,0 +1,71 @@
...
+root. Array-form schemas use the wrapper object that ESLint synthesizes around
+the positional schema array. An `id` or `$id` establishes a resource base for
+matching absolute self-references; references to other documents remain
+unresolved under ESLint's `missingRefs: 'ignore'` configuration.
+
+`schema: false` remains the explicit validation opt-out established by
+[ESLint RFC 85](https://github.com/eslint/rfcs/tree/main/designs/2021-schema-object-rules#opt-out).
+In contrast, `schema: []` accepts no options, while `schema: [{}]` carries
+array-shorthand cardinality semantics and is not classified as useless.
+
```
bmish, line 42: New dedicated section for this:

```suggestion

### Root schema forms

Use `schema: false` to explicitly disable option validation. Use `schema: []`
for a rule that accepts no options. `schema: [{}]` is different: it permits one
unconstrained option, so this rule does not report it.

```

morgan-coded:
Indexed statically declared `id`/`$id` values in the reference model: they now register draft-04 named fragments and embedded document roots, including pointers into those resources, so both shapes from your review resolve and enforce while unknown anchors and genuinely external documents still report. Each resolution row was measured against ESLint's Ajv before its fixture was pinned.

The policy rule's item scan now resolves hoisted composition branches, which also removed three `explicitItems` false positives from the core sweep — recount: core 1/6, this plugin 0/0.

Docs round applied wholesale: both intros, the one-line counterpart links, `Validator compatibility`, `Root schema forms`, and the `recommended` intent sentence, with the check count and minor-release wording gone; the PR description is refreshed to match. Local gates are green: 1,446 tests, lint, typecheck, format, generated docs, and both e2e suites.

lib/rules/no-incomplete-meta-schema.ts
```diff
@@ -0,0 +1,388 @@
...
+  isStaticallyInspectable,
+  resolveObjectExpression,
+  type ObjectProperties,
+} from './meta-schema-utils.ts';
+
+const defaultChecks = {
+  explicitAdditionalProperties: true,
+  explicitItems: true,
+  typedItems: true,
+  boundedTuples: true,
```
bmish, line 33: I found one small remaining consistency gap: hoisting a tuple array bypasses `boundedTuples` and `typedItems`:

```js
const tuple = [{ pattern: 'x' }];
schema: [{ type: 'array', items: tuple }]
```

The inline equivalent reports both checks; this form reports nothing. Since statically hoisted containers are otherwise supported, I’d fix this before merge.

docs/rules/no-useless-meta-schema.md
```diff
@@ -0,0 +1,66 @@
+# eslint-plugin/no-useless-meta-schema
+
+📝 Disallow rule options schema constructs that ESLint ignores.
+
+<!-- end auto-generated rule header -->
+
+ESLint validates rule options with a configured Ajv 6 instance. Some schema
```
bmish: Let's not lead with "ESLint validates rule options with a configured Ajv 6 instance". This rule isn't only limited to checks about ajv6, and that's too much of an implementation detail to lead with.

docs/rules/no-incorrect-meta-schema.md
bmish, line 1: Sorry about the churn but can we rename this rule to `no-incorrect-meta-schema`:

> Rename `no-useless-meta-schema` to `no-incorrect-meta-schema` because the rule covers more than useless/no-op constructs: it also reports wrong root types, unresolved references, incompatible keywords, and impossible bounds. “Incorrect” captures the full set of objective schema defects and leaves `no-incomplete-meta-schema` for valid but opinionated completeness improvements.

I'm thinking we could end up with this set of rules:

- `no-incorrect-meta-schema` — objective defects intrinsic to the schema.
- `no-incomplete-meta-schema` — opt-in, opinionated completeness improvements.
- `no-meta-schema-mismatch` — future rule for inconsistencies between the schema, defaults, and implementation - I filed #653 to track this new idea (should not be part of this PR)

bmish (approved):
Thank you!

PR #650, opened 2026-07-25
fix: string array schema validation for `meta-property-ordering` and `test-case-property-ordering` rules

### LINKED ISSUE #649: Array option schemas use `elements` instead of `items`
Hello. I came across this while looking through the plugin's rule option schemas.

Three rules use `elements` instead of `items` in their array option schema. ESLint validates rule options with JSON Schema, but `elements` is a JTD keyword rather than a JSON Schema one, so it seems ajv just ignores it and the intended item constraint isn't actually applied.

For `meta-property-ordering` and `test-case-property-ordering`, the `order` option is meant to be a string array, but passing a numeric array like `[1, 2, 3]` is accepted without an error (switching to `items` rejects it as expected):

https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/a19cec0b950444b21dcd94d2ea07969234c36d7c/lib/rules/meta-property-ordering.ts#L40

https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/a19cec0b950444b21dcd94d2ea07969234c36d7c/lib/rules/test-case-property-ordering.ts#L43

`no-property-in-node` uses the same keyword, but its case looks different: the implementation types the option as `string[]` and passes each value to `new RegExp()`, while the existing tests pass RegExp literals. So `items: { type: 'string' }` would break those tests, and the intended type isn't clear to me — probably worth a separate discussion.

https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/a19cec0b950444b21dcd94d2ea07969234c36d7c/lib/rules/no-property-in-node.ts#L69

As far as I can tell this has been there since these rules were first added, so it looks long-standing rather than a recent regression.

References:

- ESLint option schemas are based on JSON Schema — https://eslint.org/docs/latest/extend/custom-rules#options-schemas
- `elements` is a keyword from JTD (RFC 8927) — https://ajv.js.org/json-type-definition.html

Am I missing a reason for the looser validation here, or is this an oversight? If it's the latter, would it be okay for me to open a PR for the two `order` options (with a regression test)? Thank you.

electrohyun:

Bug fix. `meta-property-ordering` and `test-case-property-ordering` declared their string-array `order` option with the JTD keyword `elements` instead of the JSON Schema keyword `items`. ESLint validates options with JSON Schema (via ajv), so `elements` was silently ignored and the "each item is a string" constraint was never enforced (e.g. `["error", [1, 2, 3]]` was accepted).

Switch to `items` and add a regression test asserting non-string values are now rejected.

`no-property-in-node` uses the same mistaken keyword, but its existing tests pass regex literals for `additionalNodeTypeFiles` (the option is typed as `string[]` and each value is passed to `new RegExp()`), so switching it to `items` also requires updating those tests. That's already handled in #648, so it's intentionally left out here.

#### Related Issues

Fixes #649

The regression test uses `Linter` directly rather than `RuleTester`, because `RuleTester.run()` invokes the test framework's `describe`/`it` internally and so can't be wrapped in `expect(...).toThrow()` to assert that an invalid option is rejected.

### DIFF
--- a/lib/rules/meta-property-ordering.ts
+++ b/lib/rules/meta-property-ordering.ts
@@ -37,7 +37,7 @@ const rule: Rule.RuleModule = {
       {
         type: 'array',
         description: 'What order to enforce for meta properties.',
-        elements: { type: 'string' },
+        items: { type: 'string' },
       },
     ],
     defaultOptions: [defaultOrder],
--- a/lib/rules/test-case-property-ordering.ts
+++ b/lib/rules/test-case-property-ordering.ts
@@ -40,7 +40,7 @@ const rule: Rule.RuleModule = {
       {
         type: 'array',
         description: 'What order to enforce for test case properties.',
-        elements: { type: 'string' },
+        items: { type: 'string' },
       },
     ],
     defaultOptions: [defaultOrder],
--- a/tests/lib/schema-array-item-options.ts
+++ b/tests/lib/schema-array-item-options.ts
@@ -1,0 +1,45 @@
+import { Linter, type Rule } from 'eslint';
+import { describe, it, expect } from 'vitest';
+import metaPropertyOrdering from '../../lib/rules/meta-property-ordering.ts';
+import testCasePropertyOrdering from '../../lib/rules/test-case-property-ordering.ts';
+
+const linter = new Linter();
+
+function invalidOptionConfig(
+  ruleName: string,
+  rule: Rule.RuleModule,
+  option: unknown,
+): Linter.Config {
+  return {
+    plugins: { test: { rules: { [ruleName]: rule } } },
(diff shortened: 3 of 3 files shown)


bmish (approved):
Thanks!

lib/rules/meta-property-ordering.ts
```diff
@@ -37,7 +37,7 @@ const rule: Rule.RuleModule = {
       {
         type: 'array',
         description: 'What order to enforce for meta properties.',
-        elements: { type: 'string' },
+        items: { type: 'string' },
```
bmish, line 40: We could tighten this more later like `uniqueItems: true`.

PR #655, opened 2026-08-03
fix: port `schema`/`deprecated` in prefer-object-rule

### LINKED ISSUE #311: `prefer-object-rule` should port over potential `deprecated` and `schema` properties from function-style rule
As seen [here](https://github.com/eslint/eslint/blob/173e82040895ad53b2d9940bfb3fb67a0478f00b/tools/config-rule.js#L296), the deprecated/function-style format of ESLint rules can actually export `deprecated` and `schema` properties, and we should ensure that [eslint-plugin/prefer-object-rule](https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/prefer-object-rule.md) ports those over to the `meta` object when converting a rule to the modern object-style rule.

electrohyun:

Hello. The `prefer-object-rule` rule automatically converts function-style rules into the object style, but in this process the existing fixer only wrapped the function as `{ create(…) {} }`, and assignments like `module.exports.schema` / `module.exports.deprecated` were attached at the top level, so ESLint could not recognize them as `meta.schema` and ignored them. This PR fixes that problem. Please see issue #311 for details.

(1) I fixed the problem where the fixer only wrapped the function.

- I declared the assignments that can be moved into `meta` as `MetaAssignment`. (The reason I chose `type` instead of `interface` is that I followed the convention of the same folder.)
- The `collectMetaAssignments` function is a function that finds assignments and collects them so they can be moved into `meta`.
- It goes through `program.body` and, via an if statement, filters only the ones that satisfy the conditions (the code shape follows `getRuleExportsCJS` in `lib/utils.ts`, including its `exportsVarOverridden` handling). Then it returns the properties to port together with the statements to remove.
- After that, if there is a target to insert, it becomes the form `{meta: {…}, create …}`.
- If there is no target to insert, `metaPrefix` is `""`, so it becomes the form `{create …}`. This is exactly the same as before.
- Now the values have been organized and copied into `meta`. It removes the original `module.exports.schema`, `module.exports.deprecated` lines that were there.

(2) I added eight tests.

- It catches the function (`function`) form.
  - It catches `module.exports = function (context) { return {}; }`,
  - and moves `module.exports.schema` / `module.exports.deprecated` together,
  - and changes it into `{meta: {schema: [{ type: 'object' }], deprecated: true}, create(context) { return {}; }}`.
  - (checks that when there are multiple properties they are joined with commas inside `meta`)
- It catches the arrow function (`=>`) form.
  - It catches `module.exports = (context) => { return {}; }`,
  - and changes it into `{meta: {schema: []}, create: (context) => { return {}; }}`.
  - (also checks the case where only `schema` is present)
- `exports.<key>` is left untouched (it is not ported).
  - `exports.deprecated = true` is not moved, because in a function-style rule the `exports` binding no longer refers to the exported rule (so it is dead code),
  - and only the function is wrapped, as in `module.exports = {create(context) { return {}; }};`, leaving `exports.deprecated = true` as it was.
- Non-target keys and computed (bracket) accesses are not moved into `meta`; they are left as is.
  - `module.exports.fixable = 'code'` (a key other than schema/deprecated) or `module.exports['schema'] = []` (bracket access, not dot access) are not ported,
  - and only the function is wrapped, as in `module.exports = {create(context) { return {}; }};`, leaving those assignments as they were.
- Top-level statements unrelated to the fixer are left untouched.
  - statements like `doSomething();` are not touched,
  - and only the function is wrapped, as in `{create(context) { return {}; }}`.
- If `module.exports.meta` is already assigned, nothing is ported (so it cannot clobber the injected `meta`).
  - for `module.exports.schema = [1];` alongside `module.exports.meta = {};`, the fixer only wraps the function and leaves both assignments as they were.
- When the same property is assigned more than once, the last assignment wins and no duplicate key is produced.
  - `module.exports.schema = [1]; module.exports.schema = [2];` becomes `{meta: {schema: [2]}, create(context) { return {}; }}`, and both original lines are removed.
- An assignment placed before `module.exports = function ...` is dead code (the reassignment discards it), so it is not ported.
  - `module.exports.schema = [1];` written before `module.exports = function (context) { return {}; };` is left as it was, and only the function is wrapped.

(3) I added a section to the rule's documentation.

- The existing example (function style -> object style) is meant to explain the rule, so I left it as is, and since the `schema`/`deprecated` porting I worked on this time is a separate detailed behavior of the fixer, I added it as a new section.

#### Related Issues

fixes #311

Since this fixes the existing fixer's unintended behavior, it was written with the `fix` prefix. However, since it changes the existing behavior, I think the `feat` prefix could also be valid!

### DIFF
--- a/lib/rules/prefer-object-rule.ts
+++ b/lib/rules/prefer-object-rule.ts
@@ -8,17 +8,17 @@ import { getRuleInfo } from '../utils.ts';
 
 const META_PROPERTIES_TO_PORT = new Set(['schema', 'deprecated']);
 
-type PortableMetaAssignment = {
+type MetaAssignment = {
   key: string;
   valueNode: Expression;
 };
 
-type PortableMetaResult = {
-  properties: PortableMetaAssignment[];
+type CollectedMetaAssignments = {
+  properties: MetaAssignment[];
   statementsToRemove: Node[];
 };
 
-function getPortableMetaAssignments(program: Program): PortableMetaResult {
+function collectMetaAssignments(program: Program): CollectedMetaAssignments {
   const propertiesByKey = new Map<string, Expression>();
   const statementsToRemove: Node[] = [];
   let exportsVarOverridden = false;
@@ -84,7 +84,7 @@ function getPortableMetaAssignments(program: Program): PortableMetaResult {
 }
 
 function buildMetaPrefix(
-  properties: PortableMetaAssignment[],
+  properties: MetaAssignment[],
   sourceCode: SourceCode,
 ): string {
   if (properties.length === 0) {
@@ -131,7 +131,7 @@ const rule: Rule.RuleModule = {
           return;
         }
 
-        const { properties, statementsToRemove } = getPortableMetaAssignments(
+        const { properties, statementsToRemove } = collectMetaAssignments(
           sourceCode.ast,
         );
         const metaPrefix = buildMetaPrefix(properties, sourceCode);
--- a/lib/rules/prefer-object-rule.ts
+++ b/lib/rules/prefer-object-rule.ts
@@ -11,13 +11,17 @@ const META_PROPERTIES_TO_PORT = new Set(['schema', 'deprecated']);
 type PortableMetaAssignment = {
   key: string;
   valueNode: Expression;
-  statement: Node;
 };
 
-function getPortableMetaAssignments(
-  program: Program,
-): PortableMetaAssignment[] {
-  const results: PortableMetaAssignment[] = [];
+type PortableMetaResult = {
+  properties: PortableMetaAssignment[];
+  statementsToRemove: Node[];
+};
+
+function getPortableMetaAssignments(program: Program): PortableMetaResult {
+  const propertiesByKey = new Map<string, Expression>();
+  const statementsToRemove: Node[] = [];
+  let exportsVarOverridden = false;
   let hasExistingMetaAssignment = false;
 
   for (const statement of program.body) {
@@ -32,9 +36,22 @@ function getPortableMetaAssignments(
       continue;
     }
     const leftExpression = expression.left;
+    if (leftExpression.type !== 'MemberExpression' || leftExpression.computed) {
+      continue;
+    }
+
     if (
-      leftExpression.type !== 'MemberExpression' ||
-      leftExpression.computed ||
+      leftExpression.object.type === 'Identifier' &&
+      leftExpression.object.name === 'module' &&
+      leftExpression.property.type === 'Identifier' &&
+      leftExpression.property.name === 'exports'
+    ) {
+      exportsVarOverridden = true;
+      continue;
+    }
+
+    if (
+      !exportsVarOverridden ||
       leftExpression.object.type !== 'MemberExpression' ||
       leftExpression.object.computed ||
       leftExpression.object.object.type !== 'Identifier' ||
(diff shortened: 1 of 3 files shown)


lib/rules/prefer-object-rule.ts
bmish, line 1: Found three issues in [PR #655](https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/655):

[P2] Preserve metadata evaluation order — `lib/rules/prefer-object-rule.ts:162`

Moving the RHS to the earlier export can introduce runtime errors. For example, if `schema` is declared between the function export and `module.exports.schema = schema`, the fixed code reads it before initialization and throws a `ReferenceError`.

[P2] Parenthesize copied metadata expressions — `lib/rules/prefer-object-rule.ts:96`

`sourceCode.getText()` excludes surrounding parentheses. A valid assignment such as `module.exports.schema = (foo(), [])` generates invalid syntax: `meta: {schema: foo(), []}`.

[P2] Reset collected metadata after each export reassignment — `lib/rules/prefer-object-rule.ts:49`

The one-way `exportsVarOverridden` flag collects metadata belonging to earlier exports. If `module.exports` is reassigned, dead metadata from the prior export can be incorrectly resurrected on the final rule.

All CI checks pass, but tests lack nonliteral metadata expressions and repeated export assignments. No GitHub review or comments were posted.

bmish: Some more AI details:

<details>

### Autofix correctness

1. **Moving the RHS changes evaluation order.** For example, this is valid today:

```js
module.exports = function (context) { return {}; };
const schema = [{ type: 'object' }];
module.exports.schema = schema;
```

The proposed fix moves the `schema` read above its declaration, producing a `ReferenceError`. Side-effecting RHS expressions can likewise run in a different order relative to intervening statements.

2. **Copying AST-node text can drop required parentheses.** `sourceCode.getText()` excludes surrounding parentheses, so:

```js
module.exports.schema = (foo(), []);
```

is emitted as `meta: {schema: foo(), []}`, which is invalid syntax.

3. **Collection is not scoped to the export that produced `ruleInfo.create`.** Because `exportsVarOverridden` is one-way, a later `module.exports =` does not clear collected state. This can both resurrect metadata from an earlier, dead export onto the final rule and let `module.exports.meta` on an intermediate export suppress metadata belonging to a later clean function export.

### Suggested implementation

The safest approach may be to avoid moving RHS expressions entirely:

- Identify the exact `module.exports = ...` statement that corresponds to `ruleInfo.create`.
- Only consider property assignments after that statement and before the next direct `module.exports =` reassignment.
- Add `meta: {}` to the converted object, then rewrite `module.exports.schema` / `.deprecated` to `module.exports.meta.schema` / `.deprecated` **in place**.

Rewriting the LHS in place preserves evaluation order, parentheses, comments, duplicate-assignment side effects, and natural last-assignment-wins behavior. If the values must instead be inlined into the object literal, scope analysis alone is not sufficient because calls and assignments can still have ordering-sensitive side effects; porting would need to be restricted to demonstrably safe expressions and emitted with parentheses.

Please add regression tests for:

- A metadata value declared between the function export and property assignment.
- A parenthesized sequence expression.
- Multiple `module.exports =` assignments, including an intermediate `module.exports.meta` assignment.

One additional guard would be worthwhile: when `ruleInfo.create` resolves through an identifier to a `FunctionDeclaration`, the existing fixer already produces a broken bare `{create(...) {}}` statement; this branch also deletes the metadata assignment. Either skip fixing that shape for now or address it separately.

`fix:` looks like the right commit prefix: this restores metadata that the existing fixer currently loses rather than introducing a new rule capability.

</details>

electrohyun:
Thanks for the review. I've checked the parts I missed and fixed them.

I reworked the fixer so it no longer moves any value that can't be moved safely. Static literals are still inlined into `meta` as before, while non-static values (anything that references a binding or has side effects) are now rewritten in place, so `module.exports.schema = <value>` becomes `module.exports.meta.schema = <value>` and the value stays where it was.

This resolves the three correctness issues.

1. Since non-static values aren't moved, the evaluation order can no longer change and throw a `ReferenceError`.
2. Parentheses in expressions like `(foo(), [])` are no longer dropped, for the same reason.
3. Collection is now scoped to the `module.exports = ...` for the rule being converted and stops at the next `module.exports =`. Both cases you raised are handled. Metadata from a dead earlier export no longer leaks onto the final rule, and a `module.exports.meta` on an earlier export no longer suppresses a later clean function export.

For the `module.exports = create` shape (an identifier that resolves to a `FunctionDeclaration`), the value isn't the function node, so it falls outside the window and no metadata is ported. I ran the existing fixer on it, and the wrapping already produces `{create(...) {}}`, which doesn't parse, so that looks like a pre-existing issue rather than something introduced here. The metadata logic leaves this shape untouched, and we could address it in a follow-up PR.

I added regression tests for the three cases you asked for, plus repeated or side-effecting assignments to the same key, a mix of static and non-static assignments, and identifier-indirected exports.

I kept the `fix` prefix as you noted, since this restores lost metadata rather than adding a capability. Thanks for confirming.

The one place I went beyond your suggestion is that I still inline static literals rather than rewriting them in place too. That path only runs for static literals, so it can't cause the ordering or side-effect problems you flagged.

The trade-off is that the two forms end up mixed. A non-literal value stays as a trailing `module.exports.meta.schema = ...` line, while a literal goes inside the object.

One reason I kept the inline path is that with an empty `meta`, the output trips `require-meta-schema` from this plugin, so running autofix would leave a new error behind. Inlining literals avoids that for the common case.

If you'd rather handle every case the same way and rewrite them all in place, I can switch to that. Let me know which you'd prefer.

bmish (approved):
Thanks!

PR #657, opened 2026-08-07
fix: `getTestInfo` detects `NewExpression` and `CallExpression` depending on `ruleTesterConstructors` settings

### LINKED ISSUE #654: `getTestInfo` only recognizes `new RuleTester`
As discussed in https://github.com/typescript-eslint/typescript-eslint/issues/12628, rules using a RuleTester are only enabled if there's a `new RuleTester`, even if it is [suggested in the docs](https://eslint.org/docs/latest/integrate/nodejs-api#option-1--subclass-ruletester-and-override-run) to extend RuleTester or to use a function that creates a RuleTester, but this isn't supported by `getTestInfo`

Vinccool96:

Extends `getTestInfo` to capture `NewExpression` and `CallExpression` with name ending in `RuleTester`

`CallExpression` is now a valid option to create a `RuleTester`

Accepted class and function names end in `RuleTester` instead of only being `RuleTester`

#### Related Issues

Fixes #654 

💖


lib/utils.ts
```diff
@@ -67,11 +67,12 @@ function isNormalFunctionExpression(
...
   return (
-    node.type === 'NewExpression' &&
-    ((node.callee.type === 'Identifier' && node.callee.name === 'RuleTester') ||
+    (node.type === 'NewExpression' || node.type === 'CallExpression') &&
+    ((node.callee.type === 'Identifier' &&
+      node.callee.name.endsWith('RuleTester')) ||
       (node.callee.type === 'MemberExpression' &&
         node.callee.property.type === 'Identifier' &&
-        node.callee.property.name === 'RuleTester'))
+        node.callee.property.name.endsWith('RuleTester')))
```
StyleShit: Not sure whether `endsWith` is the correct check here
The function to create a `RuleTester` in typescript-eslint is called `createRuleTesterWithTypes`, which, IMO, is a reasonable name

Maybe we can add support for `settings['eslint-plugin'].additionalRuleTesterConstructors` or something similar instead?


(Anyway, if it's too much work, we can probably just rename the function on our end 😄)

Vinccool96: What do you think should be the check? I'm not sure I should try to parse all used methods to find what they return, nor all classes to see what they extend.

aladdin-add: I like the idea! something like:

```js
{
  settings: {
   "eslint-plugin": {
      "RuleTesterConstructors": ["RuleTester"] // the default, can be configured to other names.
    }
  }
}

aladdin-add: we just need to check the function names - ESLint is designed to check individual files; if the rule tester is defined in another file, static analysis is not possible.

Vinccool96: That is a good idea. Do I create two options, one for constructors (`ruleTester = new ClassName()`) and one for functions (`ruleTester = creatorFunction()`)?

Where in the docs do I put the info for the shared settings? Can every rule related to RuleTester override the settings?

aladdin-add (changes requested):
Thanks for updating! the changes look good to me, we just need:

* revert the change: a8e9dbdd5a415ecb3a5adf041c36446eb4670b09
* add a section in docs.

aladdin-add (approved):
LGTM, thanks!

